PR #259, opened 2026-03-31
fix: address Codex review on #256 — species inflation and collection context

jss367:
Parent PR: #256

Addresses two Codex Connect P2 review comments on #256 (always-visible browse detail panel).

### P2: Collapse species stats to one prediction per photo (`db.py`)
The original `top_species` query grouped directly on all prediction rows, so a photo with predictions from multiple models could be counted under multiple species. This inflated counts and made the "Top Species" panel disagree with the filtered photo total.

Fixed by using a CTE with `ROW_NUMBER() OVER (PARTITION BY photo_id ORDER BY confidence DESC)` to select a single best (highest-confidence, non-rejected) prediction per photo before grouping by species.

### P2: Pass collection context when requesting browse summary (`browse.html` / `app.py` / `db.py`)
`loadSummary()` only forwarded folder/rating/date/keyword filters. When a collection was active, `filterByCollection()` called `closeDetail()` which immediately triggered a summary call — but without the collection context, so the panel showed workspace-wide stats instead of the selected collection's photos.

Fixed by:
- Adding `activeCollectionId` JS state variable, set in `filterByCollection()` and cleared when switching to folder/keyword/find-similar browsing.
- Forwarding `collection_id` in `loadSummary()` query params.
- Accepting `collection_id` in the `/api/browse/summary` endpoint.
- Adding `collection_id` param to `Database.get_browse_summary()`, which builds a scoping subquery via the existing `_build_collection_query()` mechanism when a collection is active.

## Tests
284 passed, 0 failed

---

### DIFF
--- a/vireo/db.py
+++ b/vireo/db.py
@@ -1084,9 +1084,12 @@ class Database:
             parts = self._build_collection_query(collection_id)
             if parts is not None:
                 coll_folder_join, coll_join_clause, coll_where, coll_params = parts
-                # Build a subquery that returns the photo IDs in this collection
+                # Build a subquery that returns the photo IDs in this collection.
+                # Use alias "p" to match the alias expected by _build_collection_query;
+                # the subquery is wrapped in parentheses so "p" is scoped to it and
+                # does not conflict with the outer query's "p" alias.
                 coll_subquery = (
-                    f"SELECT DISTINCT p2.id FROM photos p2 "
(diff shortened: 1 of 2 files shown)


jss367:
This PR was generated by the PR Agent to address Codex review comments on #256. Please add the `claude-agent` label.

---

PR #485, opened 2026-04-10
fix: address codex review on #484

jss367:
Parent PR: #484

Addresses Codex Connect review feedback on #484.

The selection handlers (folder checkbox, thumbnail checkbox, and select-all) were updating `_previewSelected` but not triggering `checkForDuplicates()`, leaving stale duplicate badges/summary after selection changes while skip-duplicates is enabled.

Adds `checkForDuplicates()` calls in all three handlers so duplicate results are recomputed based on the current selection.

---


jss367:
Closing — triggering a full hash scan on every checkbox click would cause lag on large imports, and the duplicate check is against the DB (not other selected files), so toggling selection doesn't change which files are duplicates.

PR #504, opened 2026-04-11
fix: address codex review on #501 — revision pinning and blob purge

jss367:
Parent PR: #501

Addresses two Codex Connect P1 review comments on #501 that arrived after the earlier fix PR (#502) was merged.

### Fix 1: Pin verification hashes to an immutable HF revision (`vireo/model_verify.py`, `vireo/models.py`)

**Problem:** `_TREE_API` always fetched hashes from `main`. When upstream updates model files, previously-downloaded local files would start failing hash verification, writing `.verify_failed` and blocking pipelines as "incomplete" — even though the local files are not corrupt.

**Fix:**
- Added `fetch_repo_revision(repo)` — fetches the current HEAD commit SHA from `https://huggingface.co/api/models/{repo}`.
- `fetch_expected_hashes` now accepts an optional `revision` parameter (default `"main"`); the tree API URL uses `{revision}` instead of the literal `main`.
- `download_model` resolves the HEAD SHA before downloading and, on success, writes it to `.hf_revision` in the model directory.
- `verify_model` reads `.hf_revision` if present and passes the pinned SHA to `fetch_expected_hashes`, anchoring all future verifications to the exact commit the files were downloaded from. If network outage prevents SHA resolution, falls back to `"main"` (no `.hf_revision` written), preserving existing behavior.

### Fix 2: Purge the blob target, not only the cached snapshot path (`vireo/models.py::_purge_hf_cache_file`)

**Problem:** `try_to_load_from_cache` returns the snapshot path, which in HF's cache layout is a symlink into `blobs/`. Unlinking only the symlink left the corrupt blob on disk; `hf_hub_download` would re-link to the same bad bytes on the next retry, so all three retry attempts could fail identically.

**Fix:** Resolve the symlink with `os.path.realpath`, delete the underlying blob first, then remove the now-dangling snapshot symlink.

## Tests

- **`test_model_verify.py`**: Updated all `lambda subdir:` mock patches to also accept the new `revision` kwarg; added `test_fetch_expected_hashes_uses_revision_in_url`, `test_fetch_repo_revision_returns_sha`, `test_fetch_repo_revision_raises_on_network_failure`, `test_verify_model_uses_stored_revision`.
- **`test_models.py`**: Added `fetch_repo_revision` stubs to all download tests; added `test_download_model_writes_revision_file`; added `huggingface_hub` `sys.modules` stub to `_patch_download_model_env` so tests pass in environments without the library installed.

## Test results

**481 passed** in 22.87s

---

### DIFF
--- a/vireo/model_verify.py
+++ b/vireo/model_verify.py
@@ -14,11 +14,16 @@ import urllib.request
 from dataclasses import dataclass, field
 
 ONNX_REPO = "jss367/vireo-onnx-models"
-_TREE_API = "https://huggingface.co/api/models/{repo}/tree/main/{subdir}"
+_TREE_API = "https://huggingface.co/api/models/{repo}/tree/{revision}/{subdir}"
+_REPO_API = "https://huggingface.co/api/models/{repo}"
 _FETCH_TIMEOUT = 30  # seconds
 
 
 VERIFY_FAILED_SENTINEL = ".verify_failed"
+# Written by download_model to pin verification to the exact HF commit
+# that was downloaded. verify_model reads this to avoid false failures
+# when upstream later updates model files at the same branch tip.
+REVISION_FILE = ".hf_revision"
 
 
 class VerifyError(Exception):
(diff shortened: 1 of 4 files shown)


jss367:
Closing as superseded — the two fixes from this PR have been applied directly to `claude/model-hash-verify` in commits that landed before this branch could be rebased.

**Fix 1 (revision pinning)** and **Fix 2 (blob target purge)** from cf7c96a: already implemented on the parent branch in [587f209](https://github.com/jss367/vireo/commit/587f209). Functionally equivalent, slightly different naming (`fetch_latest_revision` vs `fetch_repo_revision`).

**Follow-up improvement from 8d61ef1** (split revision lookup from hash fetch so tree-API verification survives a repo-info outage): applied in [5a515e4](https://github.com/jss367/vireo/commit/5a515e4). My 587f209 had the same gap this commit fixed, so the improvement was a genuine net-positive and has been ported verbatim behavior-wise.

Resolving this PR's merge conflicts would have produced a mostly-empty diff (because the substance is already in the parent branch) so closing is the cleaner path. Both Codex P1 comments on #501, and the P1 on #504 itself, are now addressed directly on `claude/model-hash-verify`.

PR #534, opened 2026-04-13
fix: fall back to per-row text inference for BioCLIP 2.5 reshape error

jss367:
## Summary

- The `bioclip-2.5-vith14/text_encoder.onnx` was exported with an internal `gemm_input_reshape` node that hardcodes a single-prompt shape `{77, 1024}`. Any batch > 1 fails at runtime despite the input's declared batch dim being dynamic, so every custom-labels classify job has been silently skipping BioCLIP 2.5 and falling back to the secondary pipeline model (visible as `WARNING pipeline_job: Skipping model BioCLIP-2.5: ... Input shape:{77,80,1024}, requested shape:{77,1024}` in `~/.vireo/vireo.log`).
- `_compute_embeddings_with_progress` now calls the text encoder through `_make_text_runner`, which tries the batched call first and, on the first exception, logs a warning and switches to per-row inference for the rest of the session. This is a self-healing runtime workaround that preserves the fast path for correctly-exported models (bioclip-2, ViT-B-16) and works around the broken 2.5 export without needing to re-download weights.
- This is distinct from the earlier CoreML fix (#527) — that made the model load on CPU; this fixes a different failure at inference time.

## Test plan

- [x] New unit test `TestTextEncoderBatchFallback` uses a mock session that raises on batch > 1 (same error shape as the real ONNX error) and confirms embeddings are still produced with the correct shape.
- [x] End-to-end smoke test against the real `~/.vireo/models/bioclip-2.5-vith14/text_encoder.onnx`: batched call raises the known Reshape error, fallback produces `(5, 1024)` output.
- [x] Full suite passes: `python -m pytest tests/test_workspaces.py vireo/tests/test_db.py vireo/tests/test_app.py vireo/tests/test_photos_api.py vireo/tests/test_edits_api.py vireo/tests/test_jobs_api.py vireo/tests/test_darktable_api.py vireo/tests/test_config.py vireo/tests/test_classifier.py -v` — 452 passed.

### DIFF
--- a/vireo/classifier.py
+++ b/vireo/classifier.py
@@ -215,6 +215,38 @@ def _normalize(vec):
     return vec / norm
 
 
+def _make_text_runner(text_session, text_input_name):
+    """Return a callable that runs the text encoder on a (N, seq_len) token batch.
+
+    Some BioCLIP ONNX exports (notably bioclip-2.5-vith14) contain an internal
+    Reshape node with a hardcoded single-prompt shape, so inference fails for
+    any batch size > 1 even though the model's declared input is dynamic.
+    The returned runner tries the batched call first and, on failure, falls
+    back to per-row inference and remembers the choice for subsequent calls.
+    """
+    state = {"batched_ok": True}
+
+    def run(tokens):
+        if state["batched_ok"]:
+            try:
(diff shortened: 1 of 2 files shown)


jss367:
@codex

PR #537, opened 2026-04-13
fix: address codex review on #535

jss367:
Parent PR: #535

Addresses Codex Connect review feedback on #535.

## What changed

`pipeline_job.py`'s `classify_stage()` unconditionally called `ensure_megadetector_weights()` before the detection loop. This caused a regression in offline/air-gapped environments: reruns where every photo already has detections in the DB (so `_detect_batch` would reuse them without invoking MegaDetector at all) now aborted with a download failure instead of proceeding normally.

**Fix:** compute `_needs_detection` — true when `reclassify=True` or at least one photo is absent from `already_detected` — and only call `ensure_megadetector_weights()` when the model will actually run.

## Test results

414 passed. The 18 pre-existing failures on the PR branch are unchanged (same failures exist without this commit).

---

### DIFF
--- a/vireo/pipeline_job.py
+++ b/vireo/pipeline_job.py
@@ -963,20 +963,27 @@ def run_pipeline_job(job, runner, db_path, workspace_id, params):
 
             from datetime import datetime as dt
 
-            # Ensure MegaDetector weights are on disk before any detection runs.
-            # Failure here aborts the stage rather than falling back to full-image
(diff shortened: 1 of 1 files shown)


jss367:
Closing as redundant — the same Codex P2 fix was pushed directly to the base branch `claude/auto-download-megadetector` in commit 960d13b on #535, which is why this PR now conflicts.

PR #541, opened 2026-04-13
fix: address codex review on #535

jss367:
Parent PR: #535

Addresses Codex Connect review feedback on #535.

## What changed

`needs_fresh_detection` in both `pipeline_job.py` (`classify_stage`) and `classify_job.py` (`_detect_subjects`) was computed as:

```python
needs_fresh_detection = reclassify or any(p["id"] not in already_detected for p in photos)
```

When `photos` is empty and `reclassify=True`, this evaluates to `True` and triggers `ensure_megadetector_weights()` even though no detection pass will run (the loop body is never entered). This turned a no-op reclassify on an empty collection into a fatal network dependency — offline/air-gapped reruns would abort at startup rather than completing instantly with 0 photos processed.

**Fix:** prepend `bool(photos) and` to the predicate in both call sites so that the weight download is skipped entirely when the photo list is empty.

## Test results

432 passed.

---

### DIFF
--- a/vireo/classify_job.py
+++ b/vireo/classify_job.py
@@ -348,8 +348,10 @@ def _detect_subjects(photos, folders, runner, job, reclassify, db):
     )
 
     if detect_animals is not None and get_primary_detection is not None:
-        needs_fresh_detection = reclassify or any(
-            p["id"] not in already_detected_ids for p in photos
+        needs_fresh_detection = bool(photos) and (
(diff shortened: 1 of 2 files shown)


jss367:
Closing as redundant — the equivalent fix (with an added regression test) is already on the base branch as commit 51968ad on #535. #541's diff would be a subset of what's already merged there.

PR #557, opened 2026-04-14
fix: attach detection_id to pipeline predictions

jss367:
## Summary

- Pipeline classify stage wrote every prediction with `detection_id=NULL`, making them invisible to the workspace-scoped skip query (`get_existing_prediction_photo_ids` inner-joins on `detection_id`). Result: every pipeline run re-classified every photo.
- Fixed by threading the primary detection's id into the `_flush_batch` entry at `pipeline_job.py:1187`. When MegaDetector finds no animals, a full-image detection row is created so the FK is always valid — matching the standalone classify path in `classify_job.py:772-777`.

## Root cause

`pipeline_job.py:1187` built batch entries as:

```python
img_batch = [{"photo": photo, "folder_path": ..., "image_path": ..., "img": img}]
```

`_flush_batch` then did `entry.get("detection_id")` → `None`, which flowed into `add_prediction(detection_id=None, ...)`. Every pipeline-written row had a NULL FK.

`db.get_existing_prediction_photo_ids` scopes predictions by workspace via `JOIN detections d ON d.id = pr.detection_id WHERE d.workspace_id = ?`, and an INNER JOIN drops rows with NULL foreign keys, so the skip set was always empty for pipeline-produced predictions.

In a user DB running today: 2544/2544 predictions had NULL detection_id.

## Test plan

- [x] Added `test_pipeline_classify_stores_predictions_with_detection_id` — fails on main (writes NULL FK), passes with the fix. Covers both the with-detection and no-detection paths, and verifies the second-run skip set picks both photos up.
- [x] `python -m pytest vireo/tests/test_pipeline_job.py vireo/tests/test_classify_job.py` — 84/84 pass.
- [x] CLAUDE.md test suite (`tests/test_workspaces.py vireo/tests/test_db.py … vireo/tests/test_config.py`) — 437/437 pass.

## Note on existing orphan rows

Users with prior NULL-FK prediction rows (from earlier runs) will still see re-classification on those photos until the rows are cleaned up. Not addressed in this PR — open to a follow-up that either backfills or deletes them.

### DIFF
--- a/vireo/pipeline_job.py
+++ b/vireo/pipeline_job.py
@@ -1153,8 +1153,16 @@ def run_pipeline_job(job, runner, db_path, workspace_id, params):
                                     if emb_blob:
                                         import numpy as np
                                         embedding = np.frombuffer(emb_blob, dtype=np.float32)
+                                # _store_grouped_predictions reads
+                                # item["detection_id"] unconditionally on the
+                                # burst-grouping path (update_prediction_group_info).
+                                # Without it, a rerun with bursts crashes on the
+                                # first _existing item. pred_row carries
+                                # detection_id from get_prediction_for_photo's
+                                # predictions⋈detections join.
                                 raw_results.append({
                                     "photo": photo,
+                                    "detection_id": pred_row["detection_id"],
                                     "folder_path": folder_path,
                                     "image_path": image_path,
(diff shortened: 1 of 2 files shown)


jss367:
@codex

PR #558, opened 2026-04-14
fix: address codex review on #557

jss367:
Parent PR: #557

Addresses Codex Connect P1 review feedback on #557 (thread posted 2026-04-14T03:19:44Z).

## What changed

**`vireo/pipeline_job.py`** — Remove synthetic full-image detection creation; skip photos with no MegaDetector detection.

PR #557 inserted a `detector_model="full-image"` detection row when MegaDetector found no animals, to anchor the prediction to a detection FK. Codex correctly identified that `extract_masks_stage` calls `get_detections()` without a `detector_model` filter, so that synthetic row is treated as a real subject candidate. This bypasses the `photos_with_detections == 0` safeguard and causes SAM2 to run on full-frame boxes for photos with no detected subjects — producing misleading masks and quality scores, and hiding the "weights missing / no detections" diagnostic path.

Fix: pipeline classify is detection-driven. When `primary_det is None`, `continue` to the next photo. The `detection_id` is taken directly from `primary_det["id"]` with no else branch.

**`vireo/tests/test_pipeline_job.py`** — Update `test_pipeline_classify_stores_predictions_with_detection_id`:
- Replace assertion that `photo_without_det` IS in the skip set → assert it is NOT (correct: no detection → no classify → not in skip set)
- Add assertion that no synthetic `detector_model='full-image'` rows were inserted for the no-detection photo
- Add `MEGADETECTOR_ONNX_PATH` stub so `ensure_megadetector_weights` short-circuits without a network download (makes the test self-contained in CI)

## Test results

- `test_pipeline_classify_stores_predictions_with_detection_id` — passes
- CLAUDE.md required suite (437 tests) — all pass

---
Generated by scheduled PR Agent

PR #559, opened 2026-04-14
fix: address codex review on #557

jss367:
Parent PR: #557

Addresses Codex Connect P1 review feedback on #557.

## What changed

**`vireo/pipeline_job.py`** — remove synthetic full-image detection for photos with no MegaDetector result.

The original fix in #557 created a fake detection row (`detector_model="full-image"`) when MegaDetector found no animals, so predictions would have a valid `detection_id` FK. But `extract_masks_stage` calls `get_detections()` with no `detector_model` filter, so these synthetic rows are treated as real subject candidates:

- `photos_with_detections` increments for empty-scene photos
- The `photos_with_detections == 0` safeguard never fires when it should
- Mask extraction runs on a full-frame (0, 0, 1, 1) box, producing misleading quality scores and masking artifacts for photos with no wildlife subject

**Fix:** move the `primary_det is None` guard *before* `_prepare_image`. When MegaDetector found no detections for a photo, the pipeline classify stage skips it entirely — no image load, no synthetic detection, no prediction written. The `extract_masks_stage` diagnostic path fires correctly for all-empty collections.

**`vireo/tests/test_pipeline_job.py`** — correct the test assertions and add missing mock.

- Remove the assertion that `photo_without_det` lands in the skip set (wrong: the pipeline must not classify photos without MegaDetector detections)
- Add assertions that no synthetic detection row and no prediction are written for `photo_without_det`
- Add `ensure_megadetector_weights` mock so the test runs cleanly without `huggingface_hub` installed

## Test results

`test_pipeline_classify_stores_predictions_with_detection_id` passes. 217 tests pass in the CLAUDE.md required suite; the errors are pre-existing `ModuleNotFoundError` failures on this branch (same as reported in #557).

---

### DIFF
--- a/vireo/pipeline_job.py
+++ b/vireo/pipeline_job.py
@@ -1177,6 +1177,19 @@ def run_pipeline_job(job, runner, db_path, workspace_id, params):
                         # produced a detection, aborting classify mid-run.
                         photo_dets = det_map.get(photo["id"], [])
                         primary_det = photo_dets[0] if photo_dets else None
+
+                        # Pipeline classify only processes photos that
+                        # MegaDetector actually detected a subject in.
+                        # Synthesizing a full-image detection to anchor
+                        # the prediction FK corrupts extract_masks_stage:
+                        # get_detections() has no detector_model filter,
+                        # so a synthetic row is treated as a real subject
+                        # candidate, bypassing the photos_with_detections
+                        # == 0 safeguard and triggering mask extraction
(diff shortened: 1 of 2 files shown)


jss367:
Closing as duplicate of #558 and superseded by the direct fix on #557 (commit c34c84c). Both auto-fix PRs fired before the manual fix landed on the parent branch.

PR #566, opened 2026-04-14
Split pipeline classify into detect + per-model steps

jss367:
## Summary

The pipeline jobs view previously showed a single **Classify species** row regardless of how many models ran. With a two-model run you'd see one aggregate row with counts for *both* models, and if one model failed to load it was buried in the summary. This splits the classify phase into granular rows so each phase of work is visible in the jobs tree.

**New step tree for a classify-enabled pipeline:**

- `Detect subjects` — runs MegaDetector once across every collection photo, populating a shared detection cache
- `Classify with <Model A>` — uses cached detections; own progress bar, duration, and summary (`N predictions, K cached, F failed`)
- `Classify with <Model B>` — same, reuses the same cache
- …etc.

Previously detection was interleaved inside model 1's classify loop, and models 2+ reused the in-loop cache; now a single detect pre-pass feeds every classifier, so the detection row is honest about where that time goes.

Model load failures are now surfaced on the individual `classify:<model_id>` row (status=`failed`, error on the row) instead of being rolled into an aggregate summary, so users can see exactly which model broke.

Reclassify semantics are preserved: pre-run detection rows are snapshotted, the detect pre-pass re-runs with `reclassify=True`, and the stale rows are purged after the pre-pass only for photos whose detect iteration completed.

The pipeline wizard UI (`templates/pipeline.html`) is unchanged; its `Classify` card still lights up during the detect phase because `stages["classify"]` is flipped to `running` at the start of detection for back-compat.

## Test plan

- [x] New tests covering step_defs shape (detect + per-model rows), single-model labels, detect-before-classify ordering, one-model-fails-other-succeeds, and per-model summary counts (5 tests, all pass)
- [x] Existing multi-model resilience tests updated to assert per-model row statuses; reclassify multi-model test updated to assert the single shared detect call
- [x] Full CLAUDE.md test suite (555 pass; one pre-existing flaky `test_pipeline_auto_skips_classify_when_no_model` unrelated to this change — passes in isolation, fails only under full-suite ordering, same behavior on main)

### DIFF
--- a/vireo/pipeline_job.py
+++ b/vireo/pipeline_job.py
@@ -90,8 +90,8 @@ def _update_stages(runner, job_id, stages):
 
 def _current_phase(stages):
     """Determine the primary phase label from stage statuses."""
-    for name in ["regroup", "extract_masks", "classify", "model_loader",
-                 "previews", "thumbnails", "scan", "ingest"]:
+    for name in ["regroup", "extract_masks", "classify", "detect",
+                 "model_loader", "previews", "thumbnails", "scan", "ingest"]:
         info = stages.get(name, {})
         if info.get("status") == "running":
             return info.get("label", name)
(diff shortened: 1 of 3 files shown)


jss367:
@codex

PR #572, opened 2026-04-15
fix: drop old species keyword when encounter species is changed

jss367:
## Summary
- \`/api/encounters/species\` only **added** the new species keyword — it never removed the previously confirmed one. So if a user confirmed an encounter as "Bushtit" and later changed it to "Wrentit", every photo ended up with both keywords, and both got queued as \`keyword_add\` for the XMP sync.
- The endpoint now looks up the previous species (encounter-level \`confirmed_species\`, or burst-level \`species_override\`) from the pipeline cache **before** mutating, and if the species is changing, untags the old keyword and routes through \`_queue_keyword_remove\` — which already knows how to cancel a still-pending \`keyword_add\` (if the old species hadn't synced yet) or queue a \`keyword_remove\` (if it had).
- Replaced the hand-rolled pending-changes inserts with the existing \`_queue_keyword_add\` / \`_queue_keyword_remove\` helpers so the two operations stay symmetric.
- Added a \`previous_species\` field to the response, useful for future UI "changed from X to Y" toasts.
- Companion to #571 which handled the visual cascade from encounter to burst widgets.

### Known remaining edge case (separate from this PR)
When a burst has its own override and the user then changes the encounter-level species, the override-burst photos still keep the old encounter keyword. Fixing that cleanly needs a scope decision on whether encounter-level confirms should also touch override-burst photos (they currently do, which is a separate minor bug). Not addressing here to keep this change focused.

## Test plan
- [x] \`python -m pytest tests/test_workspaces.py vireo/tests/test_db.py vireo/tests/test_app.py vireo/tests/test_photos_api.py vireo/tests/test_edits_api.py vireo/tests/test_jobs_api.py vireo/tests/test_darktable_api.py vireo/tests/test_config.py\` — 446 passed (4 new tests)
- [x] New test: changing species before sync cancels the stale \`keyword_add\` instead of leaving a \`keyword_remove\` for an unwritten keyword.
- [x] New test: changing species after sync queues \`keyword_remove\` for the previous species.
- [x] New test: burst-level override replacement removes the previously overridden species.
- [x] New test: re-confirming the same species is a no-op (no stray \`keyword_remove\`).

### DIFF
--- a/vireo/app.py
+++ b/vireo/app.py
@@ -6004,27 +6004,46 @@ def create_app(db_path, thumb_cache_dir=None):
                 if not photo_id_set.issubset(enc_ids):
                     continue
                 target_enc = enc
-                if burst_index is not None and "bursts" in enc \
-                        and 0 <= burst_index < len(enc["bursts"]):
-                    ovr = enc["bursts"][burst_index].get("species_override")
-                    if ovr and ovr.get("species"):
-                        previous_species = ovr["species"]
-                    else:
-                        # No burst override yet — inherit the encounter's
-                        # confirmed species, which is what those photos were
-                        # actually tagged with.
-                        previous_species = enc.get("confirmed_species")
-                else:
-                    previous_species = enc.get("confirmed_species")
                 break
+
+        # If this is a burst-scoped request, the burst must actually exist in
+        # the cached encounter. A stale client (e.g. submitting after a burst
+        # regrouping) could otherwise fall through to an encounter-level cache
+        # update even though only the submitted photo_ids were retagged.
+        if burst_index is not None:
+            bursts = target_enc.get("bursts") if target_enc else None
+            if not bursts or not (0 <= burst_index < len(bursts)):
+                return json_error(
+                    f"Unknown burst_index {burst_index} for submitted photos",
+                )
+
+        if target_enc is not None:
+            if burst_index is not None:
+                ovr = target_enc["bursts"][burst_index].get("species_override")
+                if ovr and ovr.get("species"):
+                    previous_species = ovr["species"]
+                else:
+                    # No burst override yet — inherit the encounter's confirmed
+                    # species, which is what those photos were actually tagged
+                    # with.
+                    previous_species = target_enc.get("confirmed_species")
+            else:
+                previous_species = target_enc.get("confirmed_species")
 
         ws_id = db._ws_id()
 
         # If the species is changing, untag the old keyword from these photos
         # and cancel/queue a sidecar remove so the XMP stays in sync.
         if previous_species and previous_species.strip().lower() != species.lower():
+            # Match add_keyword's write path: species keywords live as root
+            # keywords (parent_id IS NULL) with is_species=1. Looking up by
+            # name alone could collide with a non-species homonym nested under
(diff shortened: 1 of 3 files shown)


jss367:
Addressed the three Codex review comments across the last two pushes:

1. **Reject invalid `burst_index`** — fixed in 37cf579. The handler now returns a 400 up-front when `burst_index` is out of range, before any DB or cache mutation. Codex re-flagged the same `else` branch at `vireo/app.py:6096` on the follow-up commit, but that branch is now only reachable when `burst_index is None` (the early-return catches the stale-client case).
2. **Constrain old-species lookup to root species keyword** — fixed in 37cf579. The `SELECT id FROM keywords` for the previous species now filters `parent_id IS NULL AND is_species = 1` to match `add_keyword`'s write path, so a nested homonym can't be selected.
3. **Record species replacement as one undoable step** — fixed in 138965b. Replacement now writes a single `species_replace` edit history row; `db._apply_undo` / `_apply_redo` swap tags atomically and symmetrically cancel-or-queue pending sidecar changes. One undo restores the previous confirmed species and its pending queue state.

The failing `codex-review` job is unrelated — the action logs show `Credit balance is too low` for the review bot's API key.

Test plan: `python -m pytest tests/test_workspaces.py vireo/tests/test_db.py vireo/tests/test_app.py vireo/tests/test_photos_api.py vireo/tests/test_edits_api.py vireo/tests/test_jobs_api.py vireo/tests/test_darktable_api.py vireo/tests/test_config.py` — 452 passed (8 new tests in this PR).

jss367:
Addressed the three Codex comments on commit 6f20644:

1. **Stale re-flag: "Reject invalid burst_index"** (line 6256) — already fixed in 37cf579. Line 6256 is the encounter-level `confirmed_species` write inside the cache-update block; that branch is only reached when `burst_index is None`, because the handler returns 400 earlier (now at line 6149-6154) for any out-of-range burst_index. No change.
2. **Verify burst photo_ids match burst_index** (line 6152) — fixed in 6e58ad7. After the in-range check, the handler now also requires submitted `photo_ids` to be a subset of `bursts[burst_index].photo_ids`, so a stale client with a still-valid index that points at a different burst can't retag unrelated photos or derive `previous_species` from the wrong override. Returns 400 with `photo_ids are not members of bursts[N]`. New test: `test_encounter_species_rejects_photo_ids_not_in_burst`.
3. **Wrap encounter species writes in a single transaction** (line 6201) — fixed in 6e58ad7. `db.untag_photo`, `db.remove_pending_changes`, and `db.record_edit` now accept `_commit=True` like the other db helpers; the `_queue_keyword_add` / `_queue_keyword_remove` helpers thread it through. The endpoint runs the entire mutation section (untag old, `add_keyword`, tag new, queue pending, `record_edit`) with `_commit=False` inside a `try: ... db.conn.commit() except: db.conn.rollback()` block. Edit-history pruning is deferred until after the commit. A mid-loop error can no longer leave half the photos retagged while history and pending-changes describe a full replacement.

Test plan: `python -m pytest tests/test_workspaces.py vireo/tests/{test_db,test_app,test_photos_api,test_edits_api,test_jobs_api,test_darktable_api,test_config}.py` — 465 passed (11 new tests in this PR).

PR #613, opened 2026-04-21
fix: surface darktable stdout in errors and follow symlinks in binary lookup

jss367:
## Summary

Fixes two adjacent bugs in `vireo/develop.py`:

- **stdout was being dropped in error messages.** darktable-cli writes its init/IO failures (`"can't init develop system"`, `"cannot find the style 'X' to apply"`) to stdout, not stderr. The old code only captured `result.stderr`, so users saw blank error messages and couldn't diagnose anything. Now both streams are combined, the non-empty one is preferred, and output is capped at 500 characters.
- **Symlinked darktable-cli binaries fail to launch on macOS.** On macOS the Homebrew cask installs darktable-cli as a symlink at `/usr/local/bin/darktable-cli` → `/Applications/darktable.app/Contents/MacOS/darktable-cli`. Invoking via the symlink dies in `dt_init` because darktable locates its bundle resources (`Resources/share/darktable/`, camera profiles, etc.) by walking up from argv[0]; under `/usr/local/bin/` that walk finds nothing. Resolving the detected path with `os.path.realpath` before handing it to subprocess fixes every invocation.

Discovered during a user-first testing session driving the Browse → Develop flow against the real app.

## Test plan
- [x] New unit tests cover the stdout-only, stderr-only, both-streams, and truncation cases
- [x] `find_darktable` resolves a tmp symlink to its real target (covers macOS Homebrew case)
- [x] CLAUDE.md smoke set: `python -m pytest tests/test_workspaces.py vireo/tests/test_db.py vireo/tests/test_app.py vireo/tests/test_photos_api.py vireo/tests/test_edits_api.py vireo/tests/test_jobs_api.py vireo/tests/test_darktable_api.py vireo/tests/test_config.py` → 545 passed

### DIFF
--- a/vireo/develop.py
+++ b/vireo/develop.py
@@ -7,6 +7,26 @@ import subprocess
 
 log = logging.getLogger(__name__)
 
+_DIAG_MAX_CHARS = 500
+
+
+def _format_subprocess_diag(stdout, stderr):
+    """Combine stdout and stderr into a single short diagnostic string.
+
+    Prefers whichever stream carries output. When both are present, labels
+    them so readers know which channel each line came from. Caps total length
+    at the last _DIAG_MAX_CHARS characters.
+    """
+    out = (stdout or "").strip()
+    err = (stderr or "").strip()
+    if out and err:
+        combined = f"stdout: {out}\nstderr: {err}"
+    else:
+        combined = out or err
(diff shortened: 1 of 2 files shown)


jss367:
@codex

PR #625, opened 2026-04-23
docs: design for new-images → pipeline source

jss367:
## Summary

Design doc for wiring the "N new images detected" banner into the pipeline so that clicking "Create a pipeline" produces a pipeline scoped to exactly the files the banner announced — not a blank wizard.

## Key decisions

- **Snapshot at click time** — freeze the new-image file list into a `new_image_snapshots` row; pipeline processes exactly that set.
- **Snapshot stores file paths, not photo IDs** — new images aren't yet in the DB.
- **Scan stage still runs** — scoped to the snapshot's parent dirs; downstream stages filter to the resolved photo IDs at the scan-to-classify seam, so the snapshot guarantee holds even if more files arrive between snapshot and run.
- **"New images" becomes a first-class Stage 1 source** — rendered only when `new_count > 0` (no disabled state).
- **Banner deep-links via `snapshot_id`** in the URL; short, scales to thousands of files.

## Test plan

- [ ] Review the design doc for anything missed (edge cases, scope creep).
- [ ] Open questions documented in the doc — any before implementation?

Once this is accepted, follow-up PR will turn the doc into an implementation plan (`2026-04-22-new-images-pipeline-plan.md`) and then execute.

### DIFF
--- a/docs/plans/2026-04-22-new-images-pipeline-design.md
+++ b/docs/plans/2026-04-22-new-images-pipeline-design.md
@@ -1,0 +1,200 @@
+# New Images → Pipeline Source
+
+## Problem
+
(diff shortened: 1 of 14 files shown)


jss367:
@codex

PR #627, opened 2026-04-23
feat: miss detection — three-category grid + pipeline integration

jss367:
## Summary

Flags photos where the photographer missed the shot and surfaces them for bulk rejection before culling.

Three categories, derived from signals already on `photos`:

- **`no_subject`** — MegaDetector confidence below threshold (no bbox)
- **`clipped_or_tiny`** — subject area < 0.5% of frame, or bbox touches edge
- **`out_of_focus`** — subject Tenengrad < `R` × background Tenengrad

Burst siblings relax the bar; singletons tighten it (standalone OOF requires both ratio and an absolute sharpness floor; standalone "tiny" needs <0.2%).

## What's in the branch

- **Data model**: four columns on `photos` (`miss_no_subject`, `miss_clipped`, `miss_oof`, `miss_computed_at`) via an additive migration.
- **Config**: six keys in `pipeline` (`miss_enabled`, `miss_det_confidence*`, `miss_bbox_area_min*`, `miss_oof_ratio`). `miss_enabled=true` by default.
- **Compute**: pure `classify_miss(row, siblings, config)` in `vireo/misses.py`; batch writer `compute_misses_for_workspace` runs as a thin pipeline stage after `regroup` (needs `burst_id`).
- **API**: `GET /api/misses` (grouped or `?category=`, with `?since=<iso-ts>`); `POST /api/misses/reject` (bulk sets `flag='reject'`); `POST /api/misses/<id>/unflag`.
- **UI**: top-level `/misses` page with three collapsible sections, bbox overlays (dashed frame for `no_subject`), per-card unflag, bulk-reject, and `J/K/U` keyboard navigation. "Review misses (N)" affordance in the pipeline-review summary bar, scoped to the current run via `?since=<min miss_computed_at>`.
- **Settings**: new "Miss Detection" section with master toggle + three sliders, mirroring the existing `reject_*` pattern.
- **Non-destructive**: bulk action sets `flag='reject'`; users purge through the existing flow. Individual false positives can be unflagged.

### Passing
- `vireo/tests/test_misses.py` — 11 unit tests covering each category × (burst, singleton) plus the None-feature safety invariant.
- `vireo/tests/test_misses_api.py` — 7 API tests (list/reject/unflag, invalid-category 400s, `since` filter).
- `vireo/tests/test_db.py` — 6 new tests (migration, list/clear/bulk-reject, rejected-exclusion, `since`).
- `vireo/tests/test_pipeline_job.py` — 69 pass, including updated `stages`-dict keys.
- `vireo/tests/test_config.py` — defaults present.
- `vireo/tests/test_userfirst_scenarios.py::test_misses_scenario` — Playwright-driven; seeds one miss per category, asserts section counts, bulk-rejects the `clipped` category (auto-accepts the confirm dialog), and verifies the count drops through `/api/misses`.

Full targeted run: `test_workspaces + test_db + test_config + test_misses + test_misses_api` → 300 passing.

### Known pre-existing flakes (not caused by this branch)
- `test_browse_lightbox_arrows_regression` and `test_browse_multiselect_shortcut_regression` time out on this machine even against pristine `main` versions of the files this branch touches — unrelated Playwright timeout variance.

## Test plan

- [x] Unit tests for `classify_miss` cover category × burst/singleton matrix.
- [x] None-feature safety — rows with `detection_conf` above threshold but NULL quality features must not be spuriously flagged (regression test added after review).
- [x] `list_misses` and `bulk_reject_miss_category` exclude photos already flagged as rejected.
- [x] API `?since=` filter restricts results by `miss_computed_at`.
- [x] Playwright scenario drives `/misses` end-to-end.
- [ ] Lightbox 1:1-zoom on click — not covered in the scenario (synthetic fixture has no on-disk image files so `/photos/<id>/full` 500s); exercised manually by `browse_lightbox.py` against `browse_seed`.

### DIFF
--- a/vireo/app.py
+++ b/vireo/app.py
@@ -2320,13 +2320,21 @@ def create_app(db_path, thumb_cache_dir=None):
     @app.route("/api/misses/reject", methods=["POST"])
     def api_misses_reject():
         """Set flag='rejected' on every photo currently flagged with the given
-        miss category. Returns {"rejected": n, "category": ...}."""
+        miss category.
+
+        Accepts an optional ``since`` ISO timestamp that mirrors the
+        ``/misses?since=...`` review-window scope; when present, only
+        photos whose miss_computed_at >= since are rejected, so the bulk
+        action matches what the user sees on screen. Returns
+        {"rejected": n, "category": ...}.
+        """
         db = _get_db()
         body = request.get_json(silent=True) or {}
         category = body.get("category")
+        since = body.get("since") or None
         if category not in ("no_subject", "clipped", "oof"):
             return jsonify({"error": "invalid category"}), 400
-        n = db.bulk_reject_miss_category(category)
+        n = db.bulk_reject_miss_category(category, since=since)
         return jsonify({"rejected": n, "category": category})
 
     @app.route("/api/misses/<int:photo_id>/unflag", methods=["POST"])
--- a/vireo/app.py
+++ b/vireo/app.py
@@ -2337,7 +2337,10 @@ def create_app(db_path, thumb_cache_dir=None):
         category = body.get("category")
         if category not in ("no_subject", "clipped", "oof"):
             return jsonify({"error": "invalid category"}), 400
-        db.clear_miss_flag(photo_id, category)
+        try:
+            db.clear_miss_flag(photo_id, category)
+        except ValueError:
+            return jsonify({"error": "photo not in active workspace"}), 404
         return jsonify({"ok": True})
 
     @app.route("/api/classify/readiness")
--- a/vireo/app.py
+++ b/vireo/app.py
@@ -2319,7 +2319,7 @@ def create_app(db_path, thumb_cache_dir=None):
 
     @app.route("/api/misses/reject", methods=["POST"])
     def api_misses_reject():
-        """Set flag='reject' on every photo currently flagged with the given
+        """Set flag='rejected' on every photo currently flagged with the given
         miss category. Returns {"rejected": n, "category": ...}."""
         db = _get_db()
         body = request.get_json(silent=True) or {}
--- a/vireo/app.py
+++ b/vireo/app.py
@@ -7454,6 +7454,7 @@ def create_app(db_path, thumb_cache_dir=None, api_token=None):
         """
         from pipeline import (
             load_photo_features,
+            load_results_raw,
             reflow,
             run_grouping,
             save_results,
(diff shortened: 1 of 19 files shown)


jss367:
🤖 Addressed Codex's latest review (commit `8da9080`):

**Pushed in `77223d1`:**
- **P2 `misses.py:176` — sub-second timestamps.** Widened `miss_computed_at` to microsecond precision so two pipeline runs finishing within the same wall-clock second don't produce identical timestamps (which would cause `?since=` and `bulk_reject_miss_category(since=…)` to leak across runs). ISO-8601 still sorts lexicographically across mixed precisions, so this is backward-compatible with existing rows.

**Already fixed in earlier commits on this branch (Codex flagged but `is_outdated=true` or addressed since):**
- P1 canonical `'rejected'` flag value
- P1 / P2 workspace scoping for `list_misses`, `clear_miss_flag`, `bulk_reject_miss_category`
- P1 read detector confidence from `detections` table
- P2 read bbox/conf from `detections` table
- P2 chunk detection-lookup IDs (`SQLITE_MAX_VARIABLE_NUMBER`)
- P1 partial-config `KeyError` guard in `classify_miss`
- P1 `since` propagated to bulk reject
- P1 skip miss stage on regroup failure
- P2 paired thresholds saved together (`miss_det_confidence_burst`, `miss_bbox_area_min_singleton`)
- P2 `misses` added to `STAGE_WEIGHTS`
- P2 `edit_history` recorded on bulk-reject
- P2 `exclude_photo_ids` honored in miss compute scope
- P2 miss stage gated on `skip_classify`

**Declining one item (P2, `settings.html:1332` — `bbox * 0.40`):**
Codex flagged that `miss_bbox_area_min_singleton = bbox * 0.40` makes singleton clipping "more permissive" than burst clipping. It's actually working as intended: the threshold is the *floor below which* a subject is flagged as too small. With singleton 0.002 < burst 0.005, a singleton must be smaller (≤0.2% of frame) than a burst photo (≤0.5%) to be flagged. The settings copy says "Singletons require an even smaller subject (40% of this) before being flagged, since without siblings a small subject may be intentional framing." That matches the PR description ("standalone 'tiny' needs <0.2%") — singletons get flagged *less* often, not more, because there's no burst-sibling context to confirm lost framing. The current derivation is correct.

Tests: targeted suite (607 passing) + `test_pipeline_job` (93 passing).

---

PR #640, opened 2026-04-24
fix(scan): drop stale derived caches on file content change

jss367:
## Summary

Fixes the thumbnail/full-image mismatch symptom reported for `_D851925.NEF` in the Feb2026 workspace (thumbnail showed a bird, lightbox showed a squirrel).

**Root cause.** When a file's bytes change on disk, the scanner updates `photos.file_hash` but leaves derived caches (thumbnail, working copy, preview pyramid) untouched. `generate_thumbnail` short-circuits when `<photo_id>.jpg` already exists, so the stale pixels stick around forever. Meanwhile `/photos/<id>/original` re-extracts the working copy on demand (1:1 zoom), so the working copy drifts to the new content while the thumbnail stays frozen on the old content — exact symptom the user reported.

**Fix.** Scanner captures the prior `file_hash` before the `UPDATE` and, if the newly computed hash differs, calls `_invalidate_derived_caches(db, vireo_dir, photo_id)`:
- deletes `thumbnails/<id>.jpg`
- deletes `working/<id>.jpg` and NULLs `working_copy_path` (so `_extract_working_copies` at the end of `scan()` rebuilds it this pass)
- deletes `previews/<id>_*.jpg`

Aligns with the "app self-heals broken state — no rm -rf fixes" convention.

## Test plan

- [x] New `test_rescan_invalidates_stale_thumbnail_when_file_content_changes` (reproduces the bird/squirrel divergence; fails without the fix)
- [x] New `test_rescan_regenerates_working_copy_when_file_content_changes` (asserts working-copy pixels reflect the new source)
- [x] `python -m pytest vireo/tests/test_scanner.py vireo/tests/test_thumbnails.py vireo/tests/test_scanner_working_copy.py -q` → 76 passed
- [x] Project default suite (`tests/test_workspaces.py`, `vireo/tests/test_db.py`, `test_app.py`, `test_photos_api.py`, `test_edits_api.py`, `test_jobs_api.py`, `test_darktable_api.py`, `test_config.py`) → 589 passed

## Notes / out of scope

- The fix is going-forward. Photos whose thumbnails drifted *before* this patch landed will only heal if their source file is touched again (scanner needs a file-hash change to fire the invalidation). For the specific `_D851925.NEF` case, a one-time `rm ~/.vireo/thumbnails/24146.jpg` will trigger regeneration from the current NEF on next view.
- Pre-existing: `file_mtime` is never updated on existing photo rows by the main loop. Not in scope here, but it means every incremental re-scan re-hashes the file until mtime matches. Worth a follow-up.

### DIFF
--- a/vireo/app.py
+++ b/vireo/app.py
@@ -3975,7 +3975,8 @@ def create_app(db_path, thumb_cache_dir=None, api_token=None):
         paths = body.get("paths", [])
         from audit import import_untracked
 
-        import_untracked(db, paths)
+        vireo_dir = os.path.dirname(app.config["THUMB_CACHE_DIR"])
+        import_untracked(db, paths, vireo_dir=vireo_dir)
         return jsonify({"ok": True, "imported": len(paths)})
 
     # -- Scan status (kept, non-job) --
--- a/vireo/app.py
+++ b/vireo/app.py
@@ -7127,7 +7127,10 @@ def create_app(db_path, thumb_cache_dir=None, api_token=None):
         active_ws = _get_db()._active_workspace_id
 
         def work(job):
-            return run_pipeline_job(job, runner, db_path, active_ws, params)
+            return run_pipeline_job(
+                job, runner, db_path, active_ws, params,
+                thumb_cache_dir=app.config["THUMB_CACHE_DIR"],
+            )
 
         job_id = runner.start(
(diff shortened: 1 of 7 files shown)


jss367:
@codex

PR #672, opened 2026-04-25
fix(new-images): drop in-flight handle on invalidation

jss367:
## Summary

Follow-up to #669 addressing the remaining Codex P2 about in-flight reuse.

`NewImagesCache.kickoff_compute` reused `_inflight[key]` regardless of generation. If `invalidate_workspaces` fired while a walk was in flight (folder/workspace change, completed scan), subsequent requests kept waiting on the obsolete event for the full duration of the stale walk — `/api/workspaces/active/new-images` returned `pending: true` while the cache key had already moved on.

## What changed

- `invalidate_workspaces` now pops `_inflight[key]` so the next `kickoff_compute` starts a fresh worker.
- The stale worker keeps running but is harmless: its `set()` call is dropped by the existing per-key generation guard, and its error write is dropped by the worker's own generation check (added in #669).
- The worker's `_inflight.pop` now does an identity check (`is event`) before deleting, so a stale worker can't trample the new worker's entry under the same key.

## Why this is a small follow-up rather than amending #669

#669 already shipped through the PR-agent's review/autofix cycle and tests are green. This race only costs one polling cycle (~3s) of stale `pending` and the fix is narrow — keeping it isolated makes the diff easy to read and the rebase trivial.

> **Base branch:** `vireo-unresponsive-debug` (will fall back to `main` automatically once #669 merges).

## Test plan

- [x] New regression test `test_invalidate_workspaces_starts_fresh_compute_after_in_flight_walk` in `test_new_images_cache.py` — fails before the fix, passes after.
- [x] All 63 new-images tests pass.
- [x] Full suite from `CLAUDE.md`: **637 passed in 16:58**.

### DIFF
--- a/vireo/new_images.py
+++ b/vireo/new_images.py
@@ -196,6 +196,16 @@ class NewImagesCache:
                 # fresh recompute via the 30s backoff window in
                 # ``kickoff_compute``.
                 self._errors.pop(key, None)
+                # Drop the in-flight handle so the next ``kickoff_compute``
+                # starts a fresh worker instead of waiting on the obsolete
+                # walk. The stale worker keeps running but its ``set`` call
+                # is dropped by the per-key generation guard, and any error
+                # it raises is dropped by the worker's own gen check. Without
+                # this, a long walk that's mid-flight when a folder/workspace
+                # change fires would force
+                # ``/api/workspaces/active/new-images`` to keep returning
+                # ``pending`` for the full duration of the stale walk.
+                self._inflight.pop(key, None)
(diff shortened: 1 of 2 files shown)


jss367:
Closing as obsolete — the parent branch superseded this with a better design (5b6db59: track (event, generation) in _inflight; b275f5d: coalesce stale-gen kickoffs via _rerun_pending). The parent keeps at most one walk per key in flight, which matters on the scan path where invalidate_workspaces fires per discovered folder — this PR would have fanned out parallel os.walk jobs.

PR #676, opened 2026-04-26
feat(new-images): surface background walk as ephemeral job

jss367:
## Summary

The new-images filesystem walk runs invisibly on every workspace open and on cache invalidation. With 25k+ unscanned files on a USB drive, the user sees no indication that anything is happening — the navbar just shows pending. This PR surfaces the walk as an ephemeral job entry in the bottom panel jobs list, with a live `N checked, M new` progress phase, and removes the entry on completion without writing to `job_history`.

- `JobRunner.start(..., ephemeral=True)` runs the job normally but skips `_persist_job` so navbar polls don't clutter the history list.
- `count_new_images_for_workspace(..., progress_callback, progress_every=250)` emits throttled progress with a guaranteed final-totals call.
- `NewImagesCache.kickoff_compute(..., on_spawn=...)` fires the hook only when a fresh worker is actually launched — cache hits, in-flight reuse, and error-backoff short-circuits never spawn a job, so steady-state polling stays quiet.
- The endpoint `api_workspace_new_images` registers the ephemeral job in `on_spawn` and plumbs progress events to it. The job's work_fn blocks on the cache event so its lifecycle mirrors the actual walk.

## Test plan

- [x] `vireo/tests/test_jobs.py` — ephemeral jobs (success + failure) skip `job_history` persistence
- [x] `vireo/tests/test_new_images.py` — `progress_callback` fires monotonically, respects `progress_every` throttling, emits final totals; existing behavior preserved when no callback is given
- [x] `vireo/tests/test_new_images_api.py` — `/api/jobs` surfaces a `new_images_walk` job during a cold walk; cache hits do not spawn additional jobs; nothing is persisted to `job_history`
- [x] Full project suite per CLAUDE.md: **637 passed in 1102s**

### DIFF
--- a/vireo/app.py
+++ b/vireo/app.py
@@ -2304,12 +2304,65 @@ def create_app(db_path, thumb_cache_dir=None, api_token=None):
         from db import Database
         from new_images import count_new_images_for_workspace
 
-        def compute():
+        def compute(progress_callback=None):
             wdb = Database(db_path)
             wdb.set_active_workspace(ws_id)
-            return count_new_images_for_workspace(wdb, ws_id)
+            return count_new_images_for_workspace(
+                wdb, ws_id, progress_callback=progress_callback,
+            )
 
-        event = cache.kickoff_compute(db_path, ws_id, compute)
+        # Surface the walk as an ephemeral job so the user can see it in the
+        # bottom panel rather than wondering why their workspace is silent.
+        # ``on_spawn`` only fires when this kickoff actually starts a new
+        # worker (cache truly cold) — cache hits and reuse of an in-flight
+        # walk skip job creation, so navbar polls don't clutter the list.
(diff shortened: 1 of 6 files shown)


jss367:
🤖 Looked at the Codex P2 (job reports `completed` while endpoint surfaces a `get_recent_error`). The fix in 3bff9ef already addresses this: it captures the cache worker's exception in a holder shared with the ephemeral job's `work_fn` and re-raises after `spawn_event.wait()`, so JobRunner marks the job `failed` with the original message. `test_api_new_images_job_marked_failed_when_walk_raises` covers it.

Verified the CLAUDE.md test suite still passes against `3bff9ef` (728 passed, ~90s). No additional push from me — would have been redundant with the existing fix.

---

PR #688, opened 2026-04-27
fix(photos): trust working copy on /original, skip RAW re-extract

jss367:
## Summary

The `/photos/{id}/original` endpoint was opening every cached working copy with PIL just to compare its pixel dimensions against the RAW sensor dims stored in `photos.width/height`, and re-extracting from RAW whenever the working copy was even slightly smaller. For Nikon NEFs (and any RAW where `rawpy.postprocess()` fails and Vireo falls back to the embedded JPEG), the embedded JPEG is a few pixels smaller than the full sensor area — so the check failed on every request and a fresh RAW decode ran each time (5–7s per photo).

This is the bug behind burst-review zoom stalls: opening a burst fires `/photos/{id}/original` for every card simultaneously; concurrent re-extractions thrash the disk and a few requests get stuck in the browser indefinitely.

The working copy *is* the canonical full-res asset (per `all pixel ops use working copy`). Serve it directly when present and skip the comparison.

## Changes

- `vireo/app.py` — in `serve_original_photo`, drop the `wc_w >= orig_w` dimension check and the per-request `PIL.Image.open` overhead. If `working_copy_path` is set and the file exists, send it.
- `vireo/tests/test_photos_api.py` — replace `test_original_endpoint_upgrades_working_copy_to_full_res` (which asserted the broken upgrade behavior) with `test_original_trusts_working_copy_even_when_smaller_than_stored_dims`, which uses an 8256×5504 working copy against stored 8280×5520 dims (the Nikon-NEF case) and asserts the response bytes equal the working-copy bytes — proving no re-extraction happened.

## Test plan

- [x] `pytest vireo/tests/test_photos_api.py -v` — 62 passed
- [x] Full required suite (`tests/test_workspaces.py vireo/tests/test_db.py vireo/tests/test_app.py vireo/tests/test_photos_api.py vireo/tests/test_edits_api.py vireo/tests/test_jobs_api.py vireo/tests/test_darktable_api.py vireo/tests/test_config.py`) — exit 0

## Notes

A separate PR is in flight to pre-generate working copies on import so the on-demand extraction path is rarely taken at all. This change is independent and lands the immediate fix for already-imported photos.

### DIFF
--- a/vireo/app.py
+++ b/vireo/app.py
@@ -9504,48 +9504,56 @@ def create_app(db_path, thumb_cache_dir=None, api_token=None):
 
         vireo_dir = os.path.dirname(app.config["THUMB_CACHE_DIR"])
 
-        # Working copy is the canonical asset *unless* it was deliberately
-        # capped below the original. The cheap signal is config + stored
-        # dims; the precise signal is the wc's actual dims on disk. Use
-        # the cheap signal first and only open the wc when capping is
-        # plausible — burst-review zoom fires this endpoint per card, so
-        # the common path must avoid PIL altogether.
+        # Decide whether to trust the working copy as the full-res asset
+        # by reading its actual on-disk dimensions, NOT the current
+        # ``working_copy_max_size`` config — the cap may have changed
+        # since the wc was generated, leaving stale capped wcs that
+        # config-based logic would misclassify as full-res.
         #
-        # Cases:
-        #   - cap <= 0 (full-res config): wc is never downsized → serve it.
-        #   - cap > 0 and orig_long <= cap: wc reflects the full source
-        #     (e.g. RAW embedded-JPEG fallback whose dims are slightly
-        #     smaller than the sensor area) → serve it.
-        #   - orig dims unknown: can't detect capping; trust the wc rather
-        #     than pay 5–7s for a speculative RAW re-extract per request.
-        #   - cap > 0 and orig_long > cap: wc *might* be capped, but might
-        #     already have been upgraded in a previous request (the
-        #     extraction below overwrites wc at full-res). Read the wc's
-        #     actual long side via the JPEG header and trust it when close
-        #     to orig_long — this both detects post-upgrade wcs and avoids
-        #     re-extract loops for RAWs whose embedded-JPEG fallback is
-        #     slightly smaller than the sensor area.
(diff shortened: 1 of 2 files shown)


jss367:
Walked through Codex's four comments against the current code at commit `70b08e0`:

**#1 (P1) and #2 (P1) — already addressed.** These describe behavior from earlier commits in this branch:
- #1's "returns any existing `working_copy_path` unconditionally" describes the original commit `3d30c00`. The current code at `vireo/app.py:9518-9556` has a non-trivial condition (compares actual on-disk wc dims to stored source dims, with a RAW-only tolerance for the embedded-JPEG-fallback case).
- #2's "`if not wc_might_be_capped` fast path uses the current `working_copy_max_size`" describes commit `14550b8`. That config-based check was replaced in `70b08e0` with on-disk dimension reading via `PIL.Image.open(...).size` (lazy, sub-ms — only reads SOF marker, no pixel decode), exactly to avoid the stale-cap-config issue Codex flagged.

**#3 (P2) — corner case, declined.** The 1% tolerance can theoretically misclassify a 4096-cap working copy as full-res when the source is 4097–4140px native. But:
- The tolerance is gated on `ext in RAW_EXTENSIONS`. Modern RAW formats are either well below 4096 (12MP phones at 4032×3024) or well above (~6000px+ for DSLR/mirrorless). The 4096–4140 band is essentially uninhabited.
- Worst-case loss when this fires: ≤44 pixels on the long side, imperceptible at 1:1 zoom.
- A more precise fix (gating tolerance on `wc_long < cap`) adds config plumbing for a corner case I can't construct from real cameras.

**#4 (P2) — deliberate trade-off, declined.** The `or not (orig_w and orig_h)` shortcut serves the wc when stored dims are unknown (legacy rows or corrupt EXIF). The alternative — re-extract from RAW — re-introduces the burst-zoom stall this PR exists to fix, *and* would itself produce a wc whose dims we'd then overwrite into `photos.width/height` (the current backfill at `serve_original_photo`'s extract path), giving us no better signal for next time. The proper fix is "always populate `photos.width/height` truthfully at scan time"; this is out of scope for this PR.

Marking this branch ready for review on the merits of #1 & #2 being stale and #3 & #4 being judgment calls. Happy to revisit either if you disagree with the reasoning.

PR #703, opened 2026-04-28
feat(duplicates): bulk-decide UI for ambiguous group buckets

jss367:
## Summary

When a duplicate scan returns hundreds of unresolved groups all sharing the same {folderA, folderB} parent-dir set (the "1000 photos in two folders" case), forcing the user to click through every group is miserable. The bulk-decide section lets them pick a folder once and resolve every group in the bucket with one click.

**Stacked on top of #701** (backend bucketing). Once #701 merges, retarget this PR to \`main\`.

### Backend

- **\`Database._apply_winner_loser_merge(winner_id, loser_ids)\`** extracted from \`apply_duplicate_resolution\` so the metadata-merge + reject path can be reused without re-running the resolver.
- **\`Database.bulk_resolve_by_folder(file_hashes, keep_folder)\`** — per hash, finds the candidate in \`keep_folder\` and forces it as winner. Surfaces skip reasons (\`no candidates\`, \`fewer than 2 candidates\`, \`no candidate in keep_folder\`) so a single bad hash doesn't poison a 1000-hash batch. Same-folder duplicates among the keep_folder candidates fall back to the deterministic resolver.
- **\`POST /api/duplicates/bulk-resolve\`** thin wrapper. Returns \`loser_ids\` so the UI can chain into \`/api/duplicates/delete-loser-files\` when the user opted in to immediate trash.

### UI

- New "Bulk decide" section above "Needs review", rendered for each bucket with \`group_count >= 2\`.
- Per bucket: folder list, example filenames, recoverable size, one **"Keep \<folder> for all N"** button per folder, plus a **"Also move extras to Trash"** toggle that triggers the chained delete after bulk-resolve succeeds.
- After successful bulk-resolve, locally prunes the resolved hashes from \`_lastScanResult\` and re-renders — avoids a full re-scan round-trip on libraries with thousands of duplicates.

## Test plan

- [x] **7 unit tests** for \`bulk_resolve_by_folder\` (single hash, multi-hash, no-candidate-in-folder skip, unknown hash skip, singleton skip, multiple-in-keep-folder fallback to resolver, metadata merge).
- [x] **3 API tests** for \`/api/duplicates/bulk-resolve\` (happy path with multi-hash batch, partial skip, input validation).
- [x] **2 Playwright E2E tests** seed a scan result with a 3-group bucket and verify (a) the "Bulk decide" section renders with the right folder buttons, (b) clicking "Keep <folder> for all N" flips DB state and removes the bucket from the page.
- [x] All 104 duplicate-related tests pass (98 backend + 4 API + 2 E2E).

## Manual verification needed before merge

- Run with a real library that has many duplicates across two folders. Confirm the bucket card readably summarizes the situation, the "Keep" buttons resolve quickly, and the "Also move extras to Trash" toggle behaves sanely on partial failures.

## Follow-ups (separate PRs)

- **PR 3c**: \`/api/folders/reveal\` endpoint + "Reveal all in Finder" button on each bucket card (reuses existing cross-platform reveal logic at \`vireo/app.py:1671\`).
- **PR 4**: Drop Rule 2 (longer-path tiebreaker) from \`resolve_duplicates\` so genuinely ambiguous groups flow into the bulk-decide UI instead of being arbitrarily auto-resolved.

### DIFF
--- a/tests/e2e/test_duplicates_bulk_decide.py
+++ b/tests/e2e/test_duplicates_bulk_decide.py
@@ -1,0 +1,117 @@
+"""E2E test: /duplicates page renders the bulk-decide section when the
+scan result includes multi-group buckets, and a single Keep-folder click
+resolves all groups in the bucket.
+"""
+import json
+
+from playwright.sync_api import expect
+
+
+def _seed_scan_with_buckets(db, folder_a, folder_b, n_groups=3):
+    """Insert a synthetic scan result with one bucket of N groups, all
+    sharing the same {folder_a, folder_b} parent-dir set."""
+    # Need real DB rows so the bulk-resolve API can find candidates.
+    a_fid = db.add_folder(folder_a)
+    b_fid = db.add_folder(folder_b)
+    proposals = []
+    file_hashes = []
+    for i in range(n_groups):
+        h = f"BULK{i}"
+        name = f"photo{i}.jpg"
+        p_a = db.add_photo(folder_id=a_fid, filename=name, extension=".jpg",
+                           file_size=1000, file_mtime=100.0, file_hash=h)
+        p_b = db.add_photo(folder_id=b_fid, filename=name, extension=".jpg",
+                           file_size=1000, file_mtime=100.0, file_hash=h)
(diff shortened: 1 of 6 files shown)


jss367:
@codex

PR #709, opened 2026-04-28
test: add pytest-timeout to fail hung tests fast

jss367:
## Summary

- Adds `pytest-timeout>=2.3` to dev deps and configures a 120s per-test timeout with thread-based cancellation in `pyproject.toml`.
- Hung tests now fail with a `Timeout` marker and stack trace instead of silently pegging CPU.

## Why

Two parallel `python -m pytest tests/...` runs (spawned by sibling Conductor workspaces following CLAUDE.md's "run tests before PR" instruction) hung at ~99% CPU for 30+ minutes today, cutting an in-progress pipeline's throughput in half (rate went from 6.1 photos/min → 3.3 photos/min on the live app, ETA jumped from ~4.6h → ~9h). Without any timeout configured, a wedged thread / `Event.wait()` / SSE generator inside a test silently burns a core forever.

`timeout_method = "thread"` matches the actual failure mode this codebase has — `signal`-based timeouts only fire if the main thread is responsive, which is exactly what's broken in the JobRunner-style hangs we see.

This doesn't fix the underlying hang in whichever test wedged today (likely something in `test_pipeline_job` / `test_scanner` / `test_new_images*`). It just makes the hang loud and bounded so we can find and fix it.

## Test plan

- [x] `pip install 'pytest-timeout>=2.3'` succeeds
- [x] `python -m pytest vireo/tests/test_config.py -q` still passes (22 passed in 4.20s)
- [x] A synthetic `time.sleep(300)` test under `--timeout=3` is killed at 3s with `+++ Timeout +++` and exit code 1

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,7 +18,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          pip install pytest pytest-cov
+          pip install pytest pytest-cov pytest-timeout
           pip install -e .
 
       - name: Run tests with coverage
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,7 +7,7 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
-    timeout-minutes: 20
+    timeout-minutes: 60
 
     steps:
       - uses: actions/checkout@v4
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -7,6 +7,7 @@ on:
 jobs:
   test:
     runs-on: ubuntu-latest
+    timeout-minutes: 20
 
     steps:
       - uses: actions/checkout@v4
@@ -26,6 +27,7 @@ jobs:
 
   lint:
     runs-on: ubuntu-latest
+    timeout-minutes: 5
 
     steps:
       - uses: actions/checkout@v4
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -50,6 +50,7 @@ dev = [
     "pytest>=9.0",
     "pytest-cov>=6.0",
     "pytest-playwright>=0.7",
+    "pytest-timeout>=2.3",
     "ruff>=0.11",
 ]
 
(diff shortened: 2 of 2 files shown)


jss367:
🤖 CI fix attempted but could not resolve the failure. Manual intervention needed.

**What I tried:**
- Reproduced the workflow setup locally on **Python 3.11** and **Python 3.14** (matching CI's `python-version: "3.14"`), with `pytest-timeout==2.4.0`, `pytest-cov`, and an editable install of the project.
- Ran `python -m pytest tests/ vireo/tests/ --timeout=120 -p no:cacheprovider` (and a 30s-timeout variant). All 2442 tests passed (22 skipped) in **6m6s on 3.14** / **6m17s on 3.11**. Slowest individual test was 12s (`test_pipeline_rerun_with_existing_prediction_and_bursts_does_not_crash`); nothing approached the 120s threshold.
- Verified `pytest-timeout` with `timeout_method = "thread"` actually fires on 3.14 with a synthetic `time.sleep(30)` test under `--timeout=3` (killed at 3s with `+++ Timeout +++` and exit 1, as expected).

**What's blocking:**
- The failing run (`actions/runs/25081746724`) ran for **45m50s** before exiting with code 1, but the CI logs require sign-in and aren't accessible to this routine. Without them I can't tell which test hung or why `pytest-timeout`'s `os._exit(1)` didn't fire at the 120s mark (a hang+kill should give ~7 min runtime, not 45 min).
- This PR's description explicitly notes it "doesn't fix the underlying hang in whichever test wedged today (likely something in `test_pipeline_job` / `test_scanner` / `test_new_images*`)" — so the failure may be the intentionally-loud manifestation of that hang. But since the local run on the same Python version doesn't reproduce it, a code change here would be speculative.

**Suggested next step for a human:**
1. Open the failed job logs at the URL above and grep for `+++ Timeout +++` / `Failed: Timeout` — that should pinpoint the hung test.
2. If no timeout marker is present, the hang may be outside a pytest test function (e.g., a daemon/subprocess that the `thread`-method timer can't interrupt before the GHA job-level `timeout-minutes` would have fired). In that case, look at the last test that started in the verbose log.
3. Once the offending test is identified, either fix its underlying hang or mark it `@pytest.mark.timeout(N)` with a tighter bound.

Stopping here per the fix-ci routine to avoid pushing a half-fix.

---

PR #732, opened 2026-05-01
extract-features: batch DINOv2 subject + global into one ORT call

jss367:
## Summary

Speeds up the extract-features stage by ~25% on Apple Silicon by issuing a single ONNX Runtime call for the DINOv2 subject + global embeddings instead of two back-to-back calls.

## Why

DINOv2 ONNX is forced onto CPU here — onnxruntime 1.24's `CoreMLExecutionProvider` crashes on the external-data layout used by the HF DINOv2 export (the workaround sits in `vireo/onnx_runtime.py:188-196`). On CPU, per-call overhead dominates: a single `session.run` with `batch=1` is ~935ms on this machine, while `batch=2` is ~1403ms — that's ~25% less wall time per image.

The pipeline runs DINOv2 twice per photo (subject crop + global proxy). Batching them into one call halves that overhead.

For 100 photos that's a couple of minutes saved; for 1000 it's a noticeable chunk. SAM2 is unaffected — it's already on CoreML and is not the bottleneck.

## What changed

- `vireo/dino_embed.py`: added `embed_batch(images, variant)`. `embed()` now delegates to it for the single-image case so there's one inference path.
- `vireo/pipeline_job.py` (extract_masks stage) and `vireo/app.py` (the standalone `/api` extract job): when both subject crop and global proxy are needed, call `embed_batch([subject_crop, proxy])` once. When `crop_subject` returns `None`, fall back to `embed(proxy)` for the single global embedding.
- `vireo/tests/test_dinov2.py`: new tests verifying `embed_batch` issues exactly one `session.run` with the right batched tensor shape, and rejects empty input.
- `vireo/tests/test_pipeline_job.py`: fixture monkeypatches `embed`/`embed_batch` instead of the now-unused `embed_subject`/`embed_global`.

## Considered and rejected

Tested loading the DINOv2 graph + sidecar in memory and handing the bytes to `InferenceSession` to bypass the CoreML path-tracking crash. It loads, but ORT reports 123 CoreML/CPU partitions (339/565 nodes supported) and marshaling between them makes inference **2.7× slower** than plain CPU (~2316ms vs ~854ms per call). CoreML is genuinely the wrong provider for this export.

## Test plan

- [x] `python -m pytest vireo/tests/test_dinov2.py vireo/tests/test_pipeline_job.py vireo/tests/test_pipeline.py vireo/tests/test_app.py vireo/tests/test_db.py vireo/tests/test_jobs_api.py tests/test_workspaces.py` — **837 passed, 1 skipped** (33 min, full suite)
- [ ] Run the full pipeline against a real folder and verify extract-features finishes successfully and embeddings still group photos sensibly

### DIFF
--- a/vireo/app.py
+++ b/vireo/app.py
@@ -9061,7 +9061,7 @@ def create_app(db_path, thumb_cache_dir=None, api_token=None):
         active_ws = _get_db()._active_workspace_id
 
         def work(job):
-            from dino_embed import embed_global, embed_subject, embedding_to_blob
+            from dino_embed import embed, embed_batch, embedding_to_blob
             from masking import (
                 crop_completeness,
                 crop_subject,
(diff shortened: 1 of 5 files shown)


jss367:
🤖 CI fix attempted: I could not reproduce the failure on commit `307a56b`.

Verification:
- `python -m pytest tests/ vireo/tests/ -n auto --cov=vireo --cov-fail-under=40` on Python 3.11 → **2495 passed, 22 skipped**, coverage 90.80%
- Same suite on Python 3.13 → **2495 passed, 22 skipped**, coverage 90.80%
- `ruff check vireo/ tests/` → all checks passed
- Targeted `vireo/tests/test_dinov2.py` + `vireo/tests/test_pipeline_job.py` → 139 passed

CI uses Python 3.14, which I don't have available locally, but nothing in the diff (`embed_batch`, `np.concatenate`, `outputs[0].astype(np.float32)`) looks 3.14-sensitive. The most likely explanation is a transient/flaky failure — please re-run the failed `test` job. If it fails again, the workflow run logs (job 73953683513) will show which test is actually failing; I couldn't fetch them through the available tools.

Not pushing any code change since there's nothing to fix.

---

PR #770, opened 2026-05-07
pipeline review: show pair photos and flag missing signals in trace

jss367:
## Summary

The algorithm-trace panel on the pipeline review page printed a bare `pair N · 32.4s · S=… kept` line and a components string where unused signals collapsed to a single dot (`subj·=…`). That hid the most important failure mode: a pair whose score is propped up by `time + meta` because one of the two photos is half-processed (no DINO embedding, no species predictions). The trace looked confident even when most of the signal was missing, and there was no way to tell *which* photos the row was about.

This PR makes the trace honest about both:

- Each trace row now shows **both filenames + tiny clickable thumbs** for the pair, so you can see at a glance whether the row is comparing two views of the same bird or — as in the case that surfaced this — a Western Bluebird on a twig vs. a (likely) Steller's Jay on a log 32 seconds later.
- For unused components, the row renders `subj=missing on DSC_1385 (w=0.35)` in muted italics instead of `subj·=0.00×0.35`. A component with weight=0 is rendered normally (no "missing" annotation), so "you turned this off" reads differently from "this photo doesn't have the data."

## Changes

**Backend — `vireo/encounters.py`**
- `compute_s_enc(... return_components=True)` now emits `missing_a` and `missing_b` per component (which photo lacks the signal). Score math is unchanged.
- Each `cut_microsegments` trace entry carries `photo_a_id`, `photo_b_id`, `photo_a_filename`, `photo_b_filename`. `segment_encounters` already forwards `enc["trace"]` unmodified.

**Frontend — `vireo/templates/pipeline_review.html`**
- New `.trace-pair-photos` block under each pair header: filename + 36×28 thumb for each photo, clickable to open the existing inspect panel. Uses the existing `pipelineResults.photos` lookup; no extra fetches.
- Components line: distinguishes missing-data from zero-weight; bare `·` marker is gone.
- Local `escapeHtml` helper added (this template didn't define one).

**Tests — `vireo/tests/test_encounters.py`**
- `test_trace_includes_pair_photo_ids_and_filenames` — covers the trace shape addition.
- `test_components_flag_which_photo_is_missing_each_signal` — covers the missing-on-which-side flagging using exactly the A-has-everything / B-has-nothing setup we hit on a real encounter.

## Follow-up (not in this PR)

The other half of the surfaced-state idea — a "compute the missing signals for these photos" action button right on the trace row — would need a new endpoint that runs the right pipeline step(s) for a photo-id list, plus a job-rollup story for it. Worth designing rather than guessing; happy to spec that as a separate PR.

## Test plan

- [x] `python -m pytest vireo/tests/test_encounters.py vireo/tests/test_pipeline.py` — 112 passed (incl. 2 new tests).
- [x] Full recommended set (`tests/test_workspaces.py vireo/tests/test_db.py vireo/tests/test_app.py vireo/tests/test_photos_api.py vireo/tests/test_edits_api.py vireo/tests/test_jobs_api.py vireo/tests/test_darktable_api.py vireo/tests/test_config.py vireo/tests/test_encounters.py vireo/tests/test_pipeline.py`) — 1022 passed, 1 skipped, 2 pre-existing keyword failures. The 2 failures pass in isolation on both clean main and this branch (pre-existing test-ordering pollution; documented in memory).
- [ ] Visual smoke: open the pipeline review page, focus an encounter that contains a half-processed photo, confirm the row shows both filenames + thumbs and the unused-component line says "missing on \<filename\> (w=0.xx)" in muted orange italics.

### DIFF
--- a/vireo/encounters.py
+++ b/vireo/encounters.py
@@ -195,13 +195,31 @@ def compute_s_enc(photo_a, photo_b, config=None, return_components=False):
     sp = sim_species(photo_a.get("species_top5"), photo_b.get("species_top5"))
     sm = sim_meta(photo_a, photo_b)
 
+    # Per-component missing flags: which photo (if any) lacks the signal.
+    # Used by the trace UI to surface "compute embeddings on photo B" hints
+    # without conflating a missing signal with an explicitly-zeroed weight.
+    has_subj_a = photo_a.get("dino_subject_embedding") is not None
+    has_subj_b = photo_b.get("dino_subject_embedding") is not None
+    has_global_a = photo_a.get("dino_global_embedding") is not None
+    has_global_b = photo_b.get("dino_global_embedding") is not None
+    has_species_a = bool(photo_a.get("species_top5"))
+    has_species_b = bool(photo_b.get("species_top5"))
+    has_time_a = ts_a is not None
+    has_time_b = ts_b is not None
+
+    missing = {
+        "time": (not has_time_a, not has_time_b),
+        "subj": (not has_subj_a, not has_subj_b),
+        "global": (not has_global_a, not has_global_b),
+        "species": (not has_species_a, not has_species_b),
+        "meta": (False, False),
+    }
+
     used = {
         "time": dt != float("inf"),
-        "subj": (photo_a.get("dino_subject_embedding") is not None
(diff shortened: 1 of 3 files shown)


jss367:
@codex

PR #778, opened 2026-05-08
Thumbnails: regenerate stale cache from photo-ID reuse

jss367:
## Summary

- Fixes the bug where the encounter grid shows the *previous* tenant's image for a photo whose ID was reused after delete (e.g., May2026 import inherited rowids 41014–41044 from a deleted Apr 28 batch, so 31 photos showed willet/etc. thumbnails while the lightbox correctly opened to the actual May 1 photo).
- `serve_thumbnail` now compares the cached file's mtime against `photos.file_mtime` and regenerates when the cache predates the source — anchoring correctness in the data, independent of whether every delete site cleans up.
- `/api/audit/remove-orphans` now routes through `db.delete_photos` + `_cleanup_cached_files_for_deleted_photos` (instead of the older `audit.remove_orphans`) so the cached thumbnails / previews / working copies don't leak when orphans are pruned. Also picks up the FK-cascading cleanup (detections, predictions, collection rules, iNat submissions) that the older path skipped.

## Why two layers

Defense in depth. `photos.id` is `INTEGER PRIMARY KEY` *without* `AUTOINCREMENT`, so SQLite reuses freed rowids — and there are still other delete paths that bypass the cache cleanup (folder consolidation in `Database.consolidate_folders`, companion-pair dedup in `scanner._merge_companion_pair`). The mtime guard makes stale cache hits structurally impossible regardless of how the row got deleted. The orphan-cleanup fix is hygiene to stop creating new leaks.

## Test plan

- [x] `vireo/tests/test_thumbnails.py` — added `test_serve_thumbnail_regenerates_when_cached_predates_source` (RED → GREEN) and `test_serve_thumbnail_serves_cached_when_thumb_newer_than_source` (regression guard).
- [x] `vireo/tests/test_audit.py` — added `test_remove_orphans_endpoint_unlinks_cached_thumbnail`.
- [x] Full project test suite passes (test_workspaces, test_db, test_app, test_photos_api, test_edits_api, test_jobs_api, test_darktable_api, test_config, test_audit, test_thumbnails, test_delete_api).

### DIFF
--- a/vireo/app.py
+++ b/vireo/app.py
@@ -9145,37 +9145,53 @@ def create_app(db_path, thumb_cache_dir=None, api_token=None):
             return "", 404
 
         db = _get_db()
-        # Collapse the existence + freshness probe into a single
-        # ``getmtime`` so a concurrent ``Clear cache`` (or parallel
-        # regeneration) that unlinks the file between two separate
-        # syscalls can't surface as a 500 to the user. ``FileNotFoundError``
-        # here is the cache-miss signal and falls through to the regen
-        # path below; nothing else should be silently swallowed, so we
-        # bind the exception narrowly.
+        # Workspace-scoped lookup hoisted before any cache mutation. A
+        # thumbnail URL only makes sense for a photo the active
+        # workspace can actually see, and the staleness branch below
+        # may unlink the cached file — that mutation must not run for
+        # a photo outside the active workspace, otherwise an ill-aimed
+        # request from workspace B can wipe workspace A's cache as a
+        # side effect. Doing this lookup once also lets the freshness
+        # check use ``photo["file_mtime"]`` directly instead of a
+        # separate SELECT. (Without this scope,
+        # ``get_canonical_image_path`` would also receive an empty
+        # folders dict for a cross-workspace photo and fall back to
+        # ``os.path.join('', photo['filename'])`` — a CWD-relative
+        # path that could read or persist a thumbnail derived from an
+        # unrelated same-named file.)
+        photo = db.get_photo(photo_id, verify_workspace=True)
+        if not photo:
+            # Stale URL — the photo was deleted, or it lives in a folder
+            # that isn't linked to this workspace. 404 is the right
+            # answer; cache pruning is the upstream fix (PR #758). We
+            # do NOT touch the cached file here — another workspace may
+            # legitimately own it.
+            return "", 404
+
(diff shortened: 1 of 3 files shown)


jss367:
Addressing the two Codex review comments on `0d2afaf`:

1. **Future-dated `file_mtime` (P2)** — already addressed in `0d2afaf` itself (the comment Codex re-posted on the new commit). After successful regeneration the route now `os.utime`s the cached file to `photos.file_mtime` (`vireo/app.py:9248`), so the staleness invariant `cached_mtime >= file_mtime` holds even when the source has a future timestamp. Regression test: `test_serve_thumbnail_does_not_loop_on_future_dated_source` (sets `file_mtime = now + 1 day`, asserts a second request does not bump the thumb mtime).

2. **`getmtime` race (P2)** — fixed in `9745f0b`. The cache-hit branch was doing `os.path.exists` then `os.path.getmtime` as separate syscalls; if the file vanished between them (concurrent `Clear cache`, parallel regeneration), `getmtime` raised `FileNotFoundError` which bubbled up as a 500. Collapsed to a single `getmtime` with `except FileNotFoundError` falling through to the self-heal regen path. Test: `test_serve_thumbnail_handles_race_between_exists_and_getmtime` (monkeypatches `getmtime` to raise once, asserts 200 not 500).

jss367:
Addressing the two new Codex comments on \`9745f0b\` in \`4cda0d7\`:

1. **Skip mtime alignment when stale wasn't rewritten (P2)** — confirmed: when \`os.remove\` failed (file lock / permissions), \`generate_thumbnail\`'s early-return-on-exists left the stale JPEG in place, and the subsequent \`os.utime\` marked it as fresh forever. Fix: on unlink failure, log a warning and \`return _send_cached\` immediately. The mtime stays untouched, so the next request retries the unlink instead of pinning the wrong image.

2. **Validate workspace before deleting cache files (P2)** — confirmed: the staleness branch unlinked the cached file before \`get_photo(verify_workspace=True)\` ran, so a cross-workspace request could destroy another workspace's cache as a side effect of returning 404. Fix: hoisted the workspace-scoped lookup to the top of the route. The freshness check now uses \`photo["file_mtime"]\` directly (one query, not two), and a cross-workspace 404 returns before any cache mutation.

Regression tests: \`test_serve_thumbnail_does_not_pin_stale_when_unlink_fails\` and \`test_serve_thumbnail_does_not_unlink_cross_workspace_cache\`.

PR #779, opened 2026-05-08
tauri: flip running app to browser mode without restart

jss367:
## Summary
- Adds **View → Open in Browser** (Cmd+Shift+B) and a tray-menu **Open in Browser** entry that work in window mode.
- Both open the live sidecar URL in the default browser and hide the WKWebView window — no restart, no port hunting.
- Reads the active port from `SidecarState`, so it works regardless of which ephemeral port the sidecar bound to. One-shot for the session; the persistent **Settings → Open in default browser on launch** preference is unchanged.

## Test plan
- [x] `cargo check` clean
- [x] Python suite (run with `--timeout=30`): 912 passed, 1 skipped
- [ ] Manually verify View → Open in Browser opens the correct port and hides the window
- [ ] Manually verify tray "Open in Browser" does the same in window mode
- [ ] Confirm Cmd+Shift+B accelerator fires

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -169,25 +169,18 @@ pub fn run() {
 
             // "Open in Browser" — flip from the WKWebView to the user's
             // default browser at runtime. Does not persist; it's a one-shot
-            // flip for this session.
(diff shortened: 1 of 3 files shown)


jss367:
@codex

PR #792, opened 2026-05-09
Pipeline review: persist Apply & Close to photo flags

jss367:
## Summary

The burst-review modal on the pipeline review page only updated the in-memory pipeline cache — picks/rejects never reached \`photos.flag\`, so a photo "rejected" in pipeline review still showed up as unflagged in Browse, Cull, and XMP sync. Exact black-box gap CORE_PHILOSOPHY warns against.

- New \`POST /api/pipeline/group/apply\` endpoint persists picks → \`flag='flagged'\`, rejects → \`flag='rejected'\`, candidates that were previously flagged → \`flag='none'\`. Adds the consensus species keyword on picks (skipped when already tagged), queues an XMP sync change, and records every change in \`edit_history\` so undo works.
- New \`POST /api/pipeline/group/state\` endpoint returns current per-photo flag + species-keyword status. The modal calls this on open so it can:
  - seed picks/rejects from the live \`photos.flag\` (not the stale pipeline cache)
  - show per-card markers for "already flagged/rejected in DB" and "already has the species keyword"
- The Apply button label now reflects the *new* writes (e.g. "Flag 2 · Reject 1 · Tag 3 as Coyote & Close") so the user sees what's about to happen.
- 9 new tests in \`vireo/tests/test_pipeline_group_apply.py\` cover all flag transitions, idempotence, keyword dedupe, edit-history recording, and validation. Full project suite (930+ tests) still passes.

## Test plan

- [x] \`python -m pytest vireo/tests/test_pipeline_group_apply.py -v\` — 9 new tests pass
- [x] Full project suite per CLAUDE.md — 930 passed, 1 skipped
- [x] \`vireo/tests/test_predictions_api.py\` + \`test_pipeline.py\` + \`test_pipeline_job.py\` — 216 passed
- [ ] **Manual** — open a burst on the pipeline review page, drag a photo to REJECT, click Apply & Close, navigate to Browse and confirm the X badge shows on that photo
- [ ] **Manual** — re-open the same burst, confirm the rejected photo lands in REJECTS automatically with the "X" pre-existing-flag marker
- [ ] **Manual** — type a consensus species, click Apply & Close on a pick, confirm the keyword shows up in Browse and the next time the burst is opened the pick has the "TAG" marker
- [ ] **Manual** — drag a previously rejected photo back to candidates, click Apply & Close, confirm Browse no longer shows it as rejected

### DIFF
--- a/vireo/app.py
+++ b/vireo/app.py
@@ -11162,7 +11162,12 @@ def create_app(db_path, thumb_cache_dir=None, api_token=None):
                     if pid in photos_with_species:
                         continue
                     db.tag_photo(pid, species_kid)
-                    db.queue_change(pid, "keyword_add", species)
+                    # Use the shared helper so a pending `keyword_remove` for
+                    # the same species cancels out instead of co-existing with
+                    # a new `keyword_add`. sync_to_xmp applies removals after
+                    # adds, so a leftover remove would strip the keyword from
+                    # XMP even though the DB still has the tag.
+                    _queue_keyword_add(pid, species)
                     kw_added_pids.append(pid)
         except ValueError as e:
             return json_error(str(e), 403)
--- a/vireo/app.py
+++ b/vireo/app.py
@@ -11056,8 +11056,16 @@ def create_app(db_path, thumb_cache_dir=None, api_token=None):
 
         species_kid = None
         if species:
+            # Read-only mirror of the species-aware lookup in db.add_keyword(
+            # species, is_species=True): match top-level taxonomy/general rows
+            # case-insensitively, prefer taxonomy. Excludes homonym rows of
+            # other deliberate types (individual, location, genre) so a person
+            # tag named like a species doesn't get reported as the species
+            # keyword for has_species_keyword / Apply-label purposes.
             row = db.conn.execute(
-                "SELECT id FROM keywords WHERE name = ? COLLATE NOCASE LIMIT 1",
+                "SELECT id FROM keywords WHERE name = ? COLLATE NOCASE "
+                "AND parent_id IS NULL AND type IN ('taxonomy', 'general') "
+                "ORDER BY (type = 'taxonomy') DESC, id ASC LIMIT 1",
                 (species,),
             ).fetchone()
             if row:
--- a/vireo/app.py
+++ b/vireo/app.py
@@ -11073,6 +11073,11 @@ def create_app(db_path, thumb_cache_dir=None, api_token=None):
 
         photos = {}
         for pid in photo_ids:
+            # Same workspace guard the apply endpoint uses, so a malicious
+            # or buggy client can't read flag/keyword state for photos that
+            # don't belong to the active workspace.
+            if not db._photo_in_workspace(pid):
+                continue
             row = db.get_photo(pid)
             if not row:
                 continue
--- a/vireo/app.py
+++ b/vireo/app.py
@@ -11037,6 +11037,151 @@ def create_app(db_path, thumb_cache_dir=None, api_token=None):
         save_results_raw(body, cache_dir, db._active_workspace_id)
         return jsonify({"ok": True})
 
+    @app.route("/api/pipeline/group/state", methods=["POST"])
+    def api_pipeline_group_state():
+        """Return current DB state for a set of photos in a pipeline burst.
+
+        Used by the burst-review modal on open so it can seed picks/rejects
+        from the live `photos.flag` value (rather than the stale cached value
+        in pipelineResults) and show whether each photo already has the
+        consensus species keyword applied.
+
+        Body: {photo_ids: [int], species: str}
+        Returns: {photos: {pid: {flag, has_species_keyword}}, species_kid: int|None}
+        """
+        db = _get_db()
+        body = request.get_json(silent=True) or {}
+        photo_ids = list(body.get("photo_ids", []) or [])
+        species = (body.get("species") or "").strip()
+
+        species_kid = None
+        if species:
+            row = db.conn.execute(
+                "SELECT id FROM keywords WHERE name = ? COLLATE NOCASE LIMIT 1",
+                (species,),
+            ).fetchone()
+            if row:
+                species_kid = row["id"]
+
+        photos = {}
+        for pid in photo_ids:
+            row = db.get_photo(pid)
+            if not row:
+                continue
+            has_kw = False
+            if species_kid is not None:
+                has_kw = any(k["id"] == species_kid for k in db.get_photo_keywords(pid))
+            photos[pid] = {
+                "flag": row["flag"] or "none",
+                "has_species_keyword": has_kw,
+            }
+        return jsonify({"photos": photos, "species_kid": species_kid})
+
+    @app.route("/api/pipeline/group/apply", methods=["POST"])
+    def api_pipeline_group_apply():
+        """Apply pick/reject decisions and species to a pipeline burst group.
+
+        Diff-based: only writes flag changes for photos whose flag actually
+        changes, only adds the species keyword when it isn't already on the
+        photo, and clears flags on photos moved to candidates that were
+        previously flagged/rejected. Returns per-photo new flag/keyword state
+        so the client can update its rendered cards without reloading.
+        """
+        db = _get_db()
+        body = request.get_json(silent=True) or {}
+        picks = list(body.get("picks", []) or [])
+        rejects = list(body.get("rejects", []) or [])
+        candidates = list(body.get("candidates", []) or [])
+        species = (body.get("species") or "").strip()
+
+        # The same photo can't be in two zones. Reject conflicting input.
+        seen = set()
+        for pid in picks + rejects + candidates:
+            if pid in seen:
+                return json_error(f"Photo {pid} appears in more than one zone", 400)
+            seen.add(pid)
+
+        # All photos must be in the active workspace.
+        for pid in picks + rejects + candidates:
+            if not db._photo_in_workspace(pid):
+                return json_error(f"Photo {pid} is not in the active workspace", 403)
+
+        # Capture current state for diffing + edit history.
+        old_flags = {}
+        for pid in picks + rejects + candidates:
+            row = db.get_photo(pid)
+            if row:
+                old_flags[pid] = row["flag"] or "none"
+
+        species_kid = None
+        photos_with_species = set()
+        if species:
+            species_kid = db.add_keyword(species, is_species=True)
+            for pid in picks:
+                kws = db.get_photo_keywords(pid)
+                if any(k["id"] == species_kid for k in kws):
+                    photos_with_species.add(pid)
+
+        try:
+            # Apply target flags. We rely on update_photo_flag's own write
+            # being a no-op at the SQL level when the value matches, but we
+            # still skip when unchanged so we don't record useless history.
+            flag_items = []
+            for pid in picks:
+                old = old_flags.get(pid, "none")
+                if old != "flagged":
+                    db.update_photo_flag(pid, "flagged")
+                    flag_items.append({"photo_id": pid, "old_value": old, "new_value": "flagged"})
+            for pid in rejects:
+                old = old_flags.get(pid, "none")
+                if old != "rejected":
+                    db.update_photo_flag(pid, "rejected")
+                    flag_items.append({"photo_id": pid, "old_value": old, "new_value": "rejected"})
+            for pid in candidates:
+                old = old_flags.get(pid, "none")
+                if old in ("flagged", "rejected"):
+                    db.update_photo_flag(pid, "none")
+                    flag_items.append({"photo_id": pid, "old_value": old, "new_value": "none"})
+
+            kw_added_pids = []
+            if species and species_kid is not None:
+                for pid in picks:
+                    if pid in photos_with_species:
+                        continue
+                    db.tag_photo(pid, species_kid)
+                    db.queue_change(pid, "keyword_add", species)
+                    kw_added_pids.append(pid)
+        except ValueError as e:
+            return json_error(str(e), 403)
+
+        if flag_items:
+            desc = (
+                f"Pipeline burst group: flagged {len(picks)}, rejected {len(rejects)}, "
+                f"cleared {sum(1 for it in flag_items if it['new_value'] == 'none')}"
+            )
+            db.record_edit("flag", desc, "pipeline_group_apply", flag_items,
+                           is_batch=len(flag_items) > 1)
+
+        if kw_added_pids and species_kid is not None:
+            kw_items = [{"photo_id": pid, "old_value": "", "new_value": str(species_kid)}
+                        for pid in kw_added_pids]
+            db.record_edit("keyword_add",
+                           f'Added "{species}" to {len(kw_added_pids)} photos (pipeline burst group)',
+                           str(species_kid), kw_items, is_batch=len(kw_added_pids) > 1)
+
+        # Return new per-photo state so the client can update without a reload.
+        result_photos = {}
+        for pid in picks + rejects + candidates:
+            row = db.get_photo(pid)
+            if not row:
+                continue
+            kws = db.get_photo_keywords(pid)
+            result_photos[pid] = {
+                "flag": row["flag"] or "none",
+                "has_species_keyword": species_kid is not None and any(k["id"] == species_kid for k in kws),
+            }
+        return jsonify({"ok": True, "photos": result_photos})
+
     @app.route("/api/pipeline/config", methods=["GET", "POST"])
     def api_pipeline_config():
         """Get or update pipeline model configuration.
--- a/vireo/templates/pipeline_review.html
+++ b/vireo/templates/pipeline_review.html
@@ -2762,7 +2762,8 @@ var grmState = {
   picks: new Set(),
   rejects: new Set(),
   removed: new Set(),
-  selected: null      // photo id
+  selected: null,     // photo id
+  sessionId: 0        // bumped on each openGroupReview; used to drop stale fetches
 };
 
 function findBurstForPhoto(photoId) {
@@ -2792,6 +2793,9 @@ function openGroupReview(encIdx, burstIdx, selectPhotoId) {
 
   var items = photoIds.map(function(pid) { return photoMap[pid]; }).filter(Boolean);
 
+  // Bump the session id so any in-flight /group/state fetch from a previous
+  // openGroupReview call lands as stale and gets ignored when it resolves.
+  var sessionId = (grmState.sessionId || 0) + 1;
   grmState = {
     encIdx: encIdx,
     burstIdx: burstIdx,
@@ -2804,7 +2808,8 @@ function openGroupReview(encIdx, burstIdx, selectPhotoId) {
     // has_species_keyword}. This is the live DB snapshot the modal opened
     // with, used to (a) render "from DB" markers and (b) compute the
     // change-only Apply button label.
-    dbState: {}
+    dbState: {},
+    sessionId: sessionId
   };
   _grmLoupeLocked = false;
   _grmZoomMultiplier = 1;
@@ -2837,8 +2842,12 @@ function openGroupReview(encIdx, burstIdx, selectPhotoId) {
     headers: { 'Content-Type': 'application/json' },
     body: JSON.stringify({ photo_ids: items.map(function(p) { return p.id; }), species: initialSpecies }),
   }, { toast: false }).then(function(data) {
+    // Drop the response if a newer openGroupReview has fired since: seeding
+    // burst A's DB state into burst B's modal would mis-flag photos on Apply.
+    if (grmState.sessionId !== sessionId) return;
     grmSeedFromDbState(data && data.photos ? data.photos : {}, items, selectPhotoId);
   }).catch(function() {
+    if (grmState.sessionId !== sessionId) return;
     // Network/error fallback: seed from heuristic only.
     grmSeedFromDbState({}, items, selectPhotoId);
   });
--- a/vireo/templates/pipeline_review.html
+++ b/vireo/templates/pipeline_review.html
@@ -2809,7 +2809,12 @@ function openGroupReview(encIdx, burstIdx, selectPhotoId) {
     // with, used to (a) render "from DB" markers and (b) compute the
     // change-only Apply button label.
     dbState: {},
-    sessionId: sessionId
+    sessionId: sessionId,
+    // False until /group/state resolves and grmSeedFromDbState runs. Apply &
+    // Close is gated on this so a click during the seed window can't ship
+    // an empty picks/rejects payload (which would clear pre-existing flags
+    // by treating every photo as a candidate).
+    seeded: false
   };
   _grmLoupeLocked = false;
   _grmZoomMultiplier = 1;
@@ -2821,6 +2826,17 @@ function openGroupReview(encIdx, burstIdx, selectPhotoId) {
   grmRefreshResetAllVisibility();
 
   document.getElementById('grmOverlay').classList.add('open');
+
+  // Disable Apply while we wait for /group/state — clicking it before
+  // seed completes would send empty picks/rejects, which the server treats
+  // as "all candidates" and silently clears prior flags.
+  var applyBtn = document.getElementById('grmApplyBtn');
+  if (applyBtn) {
+    applyBtn.disabled = true;
+    applyBtn.style.opacity = '0.5';
+    applyBtn.style.cursor = 'wait';
+    applyBtn.textContent = 'Loading…';
+  }
 
   // Reset the species input to this encounter's value so a stale value from a
   // previously-reviewed group doesn't leak in when the prior modal was closed
@@ -2902,6 +2918,16 @@ function grmSeedFromDbState(dbPhotos, items, selectPhotoId) {
     });
     initSelect = firstPick || (items[0] && items[0].id);
   }
+  // Mark the modal seeded *before* re-rendering so grmUpdateApplyLabel
+  // (called from renderGroupModal) re-enables the button.
+  grmState.seeded = true;
+  var applyBtn = document.getElementById('grmApplyBtn');
+  if (applyBtn) {
+    applyBtn.disabled = false;
+    applyBtn.style.opacity = '';
+    applyBtn.style.cursor = '';
+  }
+
   if (initSelect) {
     grmSelect(initSelect);
   } else {
@@ -3582,6 +3608,13 @@ document.addEventListener('contextmenu', function(e) {
 /* --- Apply & Close --- */
 
 async function grmApply() {
+  // Belt-and-suspenders: the button is disabled until seed completes, but
+  // a stuck-open modal whose /group/state never resolved (offline, server
+  // error after retry) must not let Apply clear flags.
+  if (!grmState.seeded) {
+    console.warn('grmApply called before /group/state seed completed; ignoring.');
+    return;
+  }
   var species = document.getElementById('grmSpecies').value.trim();
 
   var picksArr = Array.from(grmState.picks);
--- a/vireo/templates/pipeline_review.html
+++ b/vireo/templates/pipeline_review.html
@@ -2864,9 +2864,36 @@ function openGroupReview(encIdx, burstIdx, selectPhotoId) {
     grmSeedFromDbState(data && data.photos ? data.photos : {}, items, selectPhotoId);
   }).catch(function() {
     if (grmState.sessionId !== sessionId) return;
-    // Network/error fallback: seed from heuristic only.
-    grmSeedFromDbState({}, items, selectPhotoId);
+    // /group/state failed (network, 5xx, etc). Do NOT seed from an empty
+    // snapshot — that would treat every previously flagged/rejected photo
+    // as an unflagged candidate, and a click on Apply would clear those
+    // flags. Keep the modal in the "Loading…" state with Apply disabled so
+    // the user can either retry by reopening the burst or close out via ×.
+    grmShowSeedError(items, selectPhotoId);
   });
+}
+
+// /group/state failed. Render the cards with no flag state so the user can
+// still inspect photos, but leave Apply disabled and surface what's wrong
+// in the button label. We deliberately do NOT set grmState.seeded = true
+// here — grmApply early-returns until it is, so a stray click can't clear
+// flags from a degraded snapshot.
+function grmShowSeedError(items, selectPhotoId) {
+  grmState.dbState = {};
+  var initSelect = selectPhotoId || (items[0] && items[0].id);
+  if (initSelect) {
+    grmSelect(initSelect);
+  } else {
+    renderGroupModal();
+  }
+  var btn = document.getElementById('grmApplyBtn');
+  if (btn) {
+    btn.disabled = true;
+    btn.style.opacity = '0.5';
+    btn.style.cursor = 'not-allowed';
+    btn.textContent = 'Could not load — reopen to retry';
+    btn.title = 'Loading photo flag state failed; close and reopen this burst to try again.';
+  }
 }
 
 // Seed picks/rejects on modal open. Photos already flagged/rejected in the
@@ -3079,6 +3106,11 @@ function grmRefreshSpeciesKeywordState() {
 function grmUpdateApplyLabel() {
   var btn = document.getElementById('grmApplyBtn');
   if (!btn) return;
+  // While the modal is unseeded (loading or seed failed), the button owns
+  // its own label/disabled state set by openGroupReview / grmShowSeedError.
+  // Don't compute a diff against an empty/unknown dbState — it would
+  // misreport pending writes and clobber the error/loading message.
+  if (!grmState || !grmState.seeded) return;
   var diff = grmComputeDiff();
   var parts = [];
   if (diff.flagNew > 0) parts.push('Flag ' + diff.flagNew);
--- a/vireo/templates/pipeline_review.html
+++ b/vireo/templates/pipeline_review.html
@@ -1299,7 +1299,7 @@ input[type="range"]::-moz-range-thumb {
   </div>
   <div class="grm-footer">
     <label style="font-size:12px;color:var(--text-muted);">Species:</label>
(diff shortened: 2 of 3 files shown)


jss367:
🤖 Codex review addressed locally, but the routine's git proxy is returning HTTP 403 on every `git push` attempt (after the standard 2/4/8/16s retry cycle). The fix is committed locally but cannot be pushed from this routine. Patch is below — apply with `git am` against `review-burst-apply-close`.

**What this patch does** (covers the two unresolved comments on `e526e9027f`):

- **P1 [`vireo/app.py`]** — `/api/pipeline/group/apply` was calling `db.queue_change(pid, "keyword_add", species)` directly, which can leave both `keyword_remove` and `keyword_add` pending for the same photo+keyword. `sync_to_xmp` unions adds/removes and applies removals after adds, so the conflict could strip the keyword from XMP even though the DB tag was re-added. Switch to the existing `_queue_keyword_add` helper, which cancels a pending `keyword_remove` instead of stacking opposite changes.
- **P2 [`vireo/templates/pipeline_review.html`]** — modal fetched `has_species_keyword` once with the initial species, but users can edit the species field afterward, leaving TAG badges and the Apply-label diff stale. Add a debounced `oninput` refresh that re-hits `/api/pipeline/group/state` with the current species, guarded by the modal `sessionId` and a same-species re-check so a slow response from a previous species can't overwrite newer `dbState`.
- Adds a regression test (`test_apply_keyword_add_cancels_pending_keyword_remove`) that pre-queues a `keyword_remove`, runs Apply with the same species, and asserts both pending entries end up cleared (cancellation is a wash, not a swap).

Tests: 934 passed (full mandated suite + new test).

The other 4 review threads (P1 stale-response guard, P1 disable-Apply-during-seed, P2 workspace-scoping, P2 taxonomy-aware lookup) are already addressed in the current branch HEAD `e526e9027f` from the prior two commits — they only show as unresolved because Codex hasn't re-reviewed since.

```patch
From 22189d1529478501c409289b2fb44ad7790e95c4 Mon Sep 17 00:00:00 2001
From: Claude <noreply@anthropic.com>
Date: Sat, 9 May 2026 19:19:41 +0000
Subject: [PATCH] Address Codex review on pipeline burst-apply endpoint

---
 vireo/app.py                             |  7 ++-
 vireo/templates/pipeline_review.html     | 61 +++++++++++++++++++++++-
 vireo/tests/test_pipeline_group_apply.py | 58 ++++++++++++++++++++++
 3 files changed, 124 insertions(+), 2 deletions(-)

diff --git a/vireo/app.py b/vireo/app.py
index e56053d..8524e49 100644
--- a/vireo/app.py
+++ b/vireo/app.py
@@ -11162,7 +11162,12 @@ def create_app(db_path, thumb_cache_dir=None, api_token=None):
                     if pid in photos_with_species:
                         continue
                     db.tag_photo(pid, species_kid)
-                    db.queue_change(pid, "keyword_add", species)
+                    # Use the queue helper so that if a pending keyword_remove
+                    # exists for this species, it gets cancelled instead of
+                    # leaving both add and remove pending — sync_to_xmp
+                    # applies removals after adds, so the conflict would
+                    # strip the keyword from XMP despite the DB tag.
+                    _queue_keyword_add(pid, species)
                     kw_added_pids.append(pid)
         except ValueError as e:
             return json_error(str(e), 403)
diff --git a/vireo/templates/pipeline_review.html b/vireo/templates/pipeline_review.html
index 951cad4..79db174 100644
--- a/vireo/templates/pipeline_review.html
+++ b/vireo/templates/pipeline_review.html
@@ -1299,7 +1299,7 @@ input[type="range"]::-moz-range-thumb {
   </div>
   <div class="grm-footer">
     <label style="font-size:12px;color:var(--text-muted);">Species:</label>
-    <input type="text" id="grmSpecies" oninput="grmUpdateApplyLabel()" style="background:var(--bg-input);color:var(--text-primary);border:1px solid var(--border-secondary);border-radius:4px;padding:6px 10px;font-size:13px;width:200px;">
+    <input type="text" id="grmSpecies" oninput="grmOnSpeciesInput()" style="background:var(--bg-input);color:var(--text-primary);border:1px solid var(--border-secondary);border-radius:4px;padding:6px 10px;font-size:13px;width:200px;">
     <button onclick="grmRemoveFromGroup()" style="background:var(--bg-tertiary);color:var(--text-muted);border:none;border-radius:4px;padding:6px 12px;font-size:12px;cursor:pointer;">Remove from group</button>
     <div class="grm-res-control" title="Image resolution — higher = slower but better for pixel-peeping">
       <label style="font-size:12px;color:var(--text-muted);">Res:</label>
@@ -3040,6 +3040,65 @@ function grmUpdateApplyLabel() {
   btn.textContent = parts.length ? parts.join(' · ') + ' & Close' : 'Apply & Close';
 }
 
+// Re-fetch has_species_keyword from the DB when the species input changes,
+// so TAG badges and the Apply diff stay accurate after the user edits the
+// species (otherwise they'd reflect the species the modal opened with). The
+// label is updated synchronously on every keystroke; the network refresh is
+// debounced so a fast typist doesn't hammer /group/state.
+var _grmSpeciesRefreshTimer = null;
+function grmOnSpeciesInput() {
+  grmUpdateApplyLabel();
+  if (_grmSpeciesRefreshTimer) clearTimeout(_grmSpeciesRefreshTimer);
+  _grmSpeciesRefreshTimer = setTimeout(grmRefreshSpeciesKeywordState, 250);
+}
+
+function grmRefreshSpeciesKeywordState() {
+  if (!grmState || !grmState.items) return;
+  var speciesEl = document.getElementById('grmSpecies');
+  var species = speciesEl ? (speciesEl.value || '').trim() : '';
+  // Snapshot the modal session so a slow response from a prior species or a
+  // prior burst can't overwrite newer dbState (mirrors openGroupReview).
+  var sessionId = grmState.sessionId;
+
+  // Empty species → nothing tagged, just clear has_species_keyword everywhere
+  // and re-render so stale TAG badges disappear immediately.
+  if (!species) {
+    if (grmState.dbState) {
+      Object.keys(grmState.dbState).forEach(function(k) {
+        if (grmState.dbState[k]) grmState.dbState[k].has_species_keyword = false;
+      });
+    }
+    renderGroupModal();
+    grmUpdateApplyLabel();
+    return;
+  }
+
+  var pids = grmState.items.map(function(p) { return p.id; });
+  safeFetch('/api/pipeline/group/state', {
+    method: 'POST',
+    headers: { 'Content-Type': 'application/json' },
+    body: JSON.stringify({ photo_ids: pids, species: species }),
+  }, { toast: false }).then(function(data) {
+    if (!grmState || grmState.sessionId !== sessionId) return;
+    // Also bail if the user has typed a different species since we fired
+    // this request — applying stale data would mis-flag the TAG badges.
+    var nowSpecies = speciesEl ? (speciesEl.value || '').trim() : '';
+    if (nowSpecies !== species) return;
+    var photos = (data && data.photos) || {};
+    if (!grmState.dbState) grmState.dbState = {};
+    pids.forEach(function(pid) {
+      var src = photos[pid];
+      var dst = grmState.dbState[pid] || (grmState.dbState[pid] = { flag: 'none', has_species_keyword: false });
+      // Only refresh has_species_keyword — don't clobber flag, since the user
+      // may have updated it in the modal since the initial seed and the diff
+      // logic compares against the opened-with snapshot.
+      dst.has_species_keyword = !!(src && src.has_species_keyword);
+    });
+    renderGroupModal();
+    grmUpdateApplyLabel();
+  }).catch(function() { /* network error: keep stale state, no UI thrash */ });
+}
+
 // Diff the current modal zones against the DB snapshot we opened with.
 // Returns counts of new flag/reject/clear writes plus tag adds that the
 // apply call will perform.
diff --git a/vireo/tests/test_pipeline_group_apply.py b/vireo/tests/test_pipeline_group_apply.py
index f4e7a35..240e193 100644
--- a/vireo/tests/test_pipeline_group_apply.py
+++ b/vireo/tests/test_pipeline_group_apply.py
@@ -151,6 +151,64 @@ def test_apply_records_edit_history_for_undo(app_and_db):
     assert 'keyword_add' in actions
 
 
+def test_apply_keyword_add_cancels_pending_keyword_remove(app_and_db):
+    """If a previous edit queued ``keyword_remove`` for the species and the
+    user re-tags the same photo via the burst modal, Apply must cancel the
+    pending removal — not queue both — so ``sync_to_xmp`` doesn't strip the
+    keyword in XMP after the add."""
+    app, db = app_and_db
+    pid = _photo_ids(db)[0]
+    species = 'Coyote'
+
+    # Pre-state: photo has the keyword AND a pending keyword_remove for it
+    # (e.g. user removed the keyword in the regular review page earlier and
+    # hasn't synced to XMP yet).
+    kid = db.add_keyword(species, is_species=True)
+    db.tag_photo(pid, kid)
+    db.queue_change(pid, 'keyword_remove', species)
+    pending_remove = db.conn.execute(
+        "SELECT COUNT(*) AS n FROM pending_changes "
+        "WHERE photo_id = ? AND change_type = 'keyword_remove' AND value = ?",
+        (pid, species),
+    ).fetchone()['n']
+    assert pending_remove == 1
+
+    # Apply via burst modal with the same species. photos_with_species
+    # already contains pid, so no DB tag is added — but the queue write
+    # path must still go through _queue_keyword_add to handle the case
+    # where the keyword was untagged-but-pending-removal. To exercise the
+    # cancellation path explicitly, untag first so the apply re-tags.
+    db.untag_photo(pid, kid)
+
+    client = app.test_client()
+    resp = client.post('/api/pipeline/group/apply', json={
+        'picks': [pid],
+        'rejects': [],
+        'candidates': [],
+        'species': species,
+    })
+    assert resp.status_code == 200
+
+    # The pending keyword_remove must be gone, AND no keyword_add should
+    # have been queued (the cancellation is a wash, not a swap).
+    pending_remove_after = db.conn.execute(
+        "SELECT COUNT(*) AS n FROM pending_changes "
+        "WHERE photo_id = ? AND change_type = 'keyword_remove' AND value = ?",
+        (pid, species),
+    ).fetchone()['n']
+    pending_add_after = db.conn.execute(
+        "SELECT COUNT(*) AS n FROM pending_changes "
+        "WHERE photo_id = ? AND change_type = 'keyword_add' AND value = ?",
+        (pid, species),
+    ).fetchone()['n']
+    assert pending_remove_after == 0
+    assert pending_add_after == 0
+
+    # And the keyword should be back on the photo.
+    kws = {k['name'] for k in db.get_photo_keywords(pid)}
+    assert species in kws
+
+
 def test_apply_rejects_conflicting_zones(app_and_db):
     app, db = app_and_db
     pids = _photo_ids(db)
-- 
2.43.0
```


---

jss367:
CI fix attempted but I could not isolate the failing test from workflow run [25609082745](https://github.com/jss367/vireo/actions/runs/25609082745) — manual intervention needed.

What I tried:
- Pulled `review-burst-apply-close` (head `ebf97e8`), installed deps, ran the full CI command locally:
  `python -m pytest tests/ vireo/tests/ -n auto -v --tb=short --cov=vireo --cov-report=term-missing --cov-fail-under=40`
- Ran `vireo/tests/test_pipeline_group_apply.py` (the new file in this PR) in isolation and under `-n auto` — all 10 tests pass.
- Ran `ruff check vireo/ tests/` — clean (also passed in CI).

What I found locally:
- 3 failures, all in scanner/pipeline-scan tests that rely on `chmod 000` actually denying reads:
  - `vireo/tests/test_pipeline_job.py::test_pipeline_scan_surfaces_permission_denied`
  - `vireo/tests/test_scanner.py::test_scan_surfaces_permission_denied_subdirs`
  - `vireo/tests/test_scanner.py::test_scan_continues_after_permission_denied`
  These are root-only artifacts (root bypasses Unix permission bits) and won't reproduce on the CI `runner` user — so they're not the cause of the CI failure.
- One transient failure in `test_extract_masks_stage_warns_on_mixed_already_masked_and_subthreshold` that did not reproduce when re-run in isolation or in a smaller xdist batch — looks flaky.

What's blocking me:
- The job logs at `actions/jobs/75176003980/logs` are auth-gated and I have no token. The `check-runs/.../annotations` endpoint also returns 403, and the rendered job page only loads logs via authenticated JS. Without the actual `FAILED ... AssertionError` line(s), I can't pin down which test broke on CI.
- The PR's diff only touches `vireo/app.py` (two new endpoints), `vireo/templates/pipeline_review.html` (JS), and the new `test_pipeline_group_apply.py` — nothing in scanner/mask-extract code paths.

Suggested next step: re-run the failed test job. If it passes, this was a flake (most likely `test_extract_masks_stage_warns_on_mixed_already_masked_and_subthreshold`, which I observed flaking locally). If it fails again, please paste the FAILED line from the log here and I can address it.

---

jss367:
Following up on the CI-fix comment above: no action needed.

- The failing run [25609082745](https://github.com/jss367/vireo/actions/runs/25609082745) was against `ebf97e8`. Commit `e526e90` was pushed afterward (the "Address remaining Codex review feedback" commit), and the next `test` job — [run 25609329367](https://github.com/jss367/vireo/actions/runs/25609329367/job/75176671329) — passed on the current HEAD. `lint` is also green and `mergeable_state` is `clean`.
- Net effect: the previous run was a flake (matches the local-repro observation that `test_extract_masks_stage_warns_on_mixed_already_masked_and_subthreshold` was the only non-permission-bit failure and didn't reproduce in isolation). No code change is required for the CI failure.
- Heads-up: the patch in [comment 4413472754](https://github.com/jss367/vireo/pull/792#issuecomment-4413472754) (`_queue_keyword_add` swap + species-input refresh + new `test_apply_keyword_add_cancels_pending_keyword_remove`) is **not** in HEAD — that prior routine reported a `git push` 403 and never landed it. That's a separate Codex-review task, not in scope for this comment, but worth flagging so it doesn't get lost.

---

jss367:
🤖 Same git-proxy 403 as the previous attempt — push from this routine is still blocked. The fix is committed locally on `review-burst-apply-close` (parent `e526e9027f`); HTTP 403 returned on every `git push` even after the standard 2/4/8/16s retry cycle.

Verbose curl shows it's the receive-pack response, not auth/credential negotiation:
```
=> Send header: Content-Type: application/x-git-receive-pack-request
<= Recv header: HTTP/1.1 403 Forbidden
<= Recv header: Server-Timing: x-originResponse;dur=1120
```

**What this patch does** (covers the two unresolved comments on `e526e9027f`):

- **P1 [`vireo/app.py`]** — `/api/pipeline/group/apply` was calling `db.queue_change(pid, "keyword_add", species)` directly, which can leave both `keyword_remove` and `keyword_add` pending for the same photo+keyword. `sync_to_xmp` unions adds/removes and applies removals after adds, so the conflict could strip the keyword from XMP even though the DB tag was re-added. Switch to the existing `_queue_keyword_add` helper, which cancels a pending `keyword_remove` instead of stacking opposite changes.
- **P2 [`vireo/templates/pipeline_review.html`]** — modal fetched `has_species_keyword` once with the initial species, but users can edit the species field afterward, leaving TAG badges and the Apply-label diff stale. Add a debounced `oninput` refresh that re-hits `/api/pipeline/group/state` with the current species, guarded by the modal `sessionId` and a same-species re-check so a slow response from a previous species can't overwrite newer `dbState`.
- Adds a regression test (`test_apply_keyword_add_cancels_pending_keyword_remove`) that pre-queues a `keyword_remove`, runs Apply with the same species, and asserts both pending entries end up cleared (cancellation is a wash, not a swap).

Tests: 922 passed (full mandated suite from CLAUDE.md) + 12 passed (`vireo/tests/test_pipeline_group_apply.py`, including the new regression test) = 934. Ruff clean on the Python files.

The other 4 review threads (P1 stale-response guard, P1 disable-Apply-during-seed, P2 workspace-scoping, P2 taxonomy-aware lookup) are already addressed in the current branch HEAD `e526e9027f` from the prior two commits — they only show as unresolved because Codex hasn't re-reviewed since.

Apply with `git am` against `review-burst-apply-close`:

```patch
From 7551a50bddb0702eaa01e2d164c589926c5ea295 Mon Sep 17 00:00:00 2001
From: Claude <noreply@anthropic.com>
Date: Sat, 9 May 2026 19:28:50 +0000
Subject: [PATCH] Address Codex review on pipeline burst-apply endpoint

---
 vireo/app.py                             |  7 ++-
 vireo/templates/pipeline_review.html     | 61 +++++++++++++++++++++++-
 vireo/tests/test_pipeline_group_apply.py | 53 ++++++++++++++++++++
 3 files changed, 119 insertions(+), 2 deletions(-)

diff --git a/vireo/app.py b/vireo/app.py
index e56053d..96fea63 100644
--- a/vireo/app.py
+++ b/vireo/app.py
@@ -11162,7 +11162,12 @@ def create_app(db_path, thumb_cache_dir=None, api_token=None):
                     if pid in photos_with_species:
                         continue
                     db.tag_photo(pid, species_kid)
-                    db.queue_change(pid, "keyword_add", species)
+                    # Use _queue_keyword_add so a pending keyword_remove for
+                    # this species gets cancelled instead of leaving both add
+                    # and remove queued — sync_to_xmp applies removals after
+                    # adds, so the conflict would strip the keyword from XMP
+                    # despite the DB tag being re-added.
+                    _queue_keyword_add(pid, species)
                     kw_added_pids.append(pid)
         except ValueError as e:
             return json_error(str(e), 403)
diff --git a/vireo/templates/pipeline_review.html b/vireo/templates/pipeline_review.html
index 951cad4..79db174 100644
--- a/vireo/templates/pipeline_review.html
+++ b/vireo/templates/pipeline_review.html
@@ -1299,7 +1299,7 @@ input[type="range"]::-moz-range-thumb {
   </div>
   <div class="grm-footer">
     <label style="font-size:12px;color:var(--text-muted);">Species:</label>
-    <input type="text" id="grmSpecies" oninput="grmUpdateApplyLabel()" style="background:var(--bg-input);color:var(--text-primary);border:1px solid var(--border-secondary);border-radius:4px;padding:6px 10px;font-size:13px;width:200px;">
+    <input type="text" id="grmSpecies" oninput="grmOnSpeciesInput()" style="background:var(--bg-input);color:var(--text-primary);border:1px solid var(--border-secondary);border-radius:4px;padding:6px 10px;font-size:13px;width:200px;">
     <button onclick="grmRemoveFromGroup()" style="background:var(--bg-tertiary);color:var(--text-muted);border:none;border-radius:4px;padding:6px 12px;font-size:12px;cursor:pointer;">Remove from group</button>
     <div class="grm-res-control" title="Image resolution — higher = slower but better for pixel-peeping">
       <label style="font-size:12px;color:var(--text-muted);">Res:</label>
@@ -3040,6 +3040,65 @@ function grmUpdateApplyLabel() {
   btn.textContent = parts.length ? parts.join(' · ') + ' & Close' : 'Apply & Close';
 }
 
+// Re-fetch has_species_keyword from the DB when the species input changes,
+// so TAG badges and the Apply diff stay accurate after the user edits the
+// species (otherwise they'd reflect the species the modal opened with). The
+// label is updated synchronously on every keystroke; the network refresh is
+// debounced so a fast typist doesn't hammer /group/state.
+var _grmSpeciesRefreshTimer = null;
+function grmOnSpeciesInput() {
+  grmUpdateApplyLabel();
+  if (_grmSpeciesRefreshTimer) clearTimeout(_grmSpeciesRefreshTimer);
+  _grmSpeciesRefreshTimer = setTimeout(grmRefreshSpeciesKeywordState, 250);
+}
+
+function grmRefreshSpeciesKeywordState() {
+  if (!grmState || !grmState.items) return;
+  var speciesEl = document.getElementById('grmSpecies');
+  var species = speciesEl ? (speciesEl.value || '').trim() : '';
+  // Snapshot the modal session so a slow response from a prior species or a
+  // prior burst can't overwrite newer dbState (mirrors openGroupReview).
+  var sessionId = grmState.sessionId;
+
+  // Empty species → nothing tagged, just clear has_species_keyword everywhere
+  // and re-render so stale TAG badges disappear immediately.
+  if (!species) {
+    if (grmState.dbState) {
+      Object.keys(grmState.dbState).forEach(function(k) {
+        if (grmState.dbState[k]) grmState.dbState[k].has_species_keyword = false;
+      });
+    }
+    renderGroupModal();
+    grmUpdateApplyLabel();
+    return;
+  }
+
+  var pids = grmState.items.map(function(p) { return p.id; });
+  safeFetch('/api/pipeline/group/state', {
+    method: 'POST',
+    headers: { 'Content-Type': 'application/json' },
+    body: JSON.stringify({ photo_ids: pids, species: species }),
+  }, { toast: false }).then(function(data) {
+    if (!grmState || grmState.sessionId !== sessionId) return;
+    // Also bail if the user has typed a different species since we fired
+    // this request — applying stale data would mis-flag the TAG badges.
+    var nowSpecies = speciesEl ? (speciesEl.value || '').trim() : '';
+    if (nowSpecies !== species) return;
+    var photos = (data && data.photos) || {};
+    if (!grmState.dbState) grmState.dbState = {};
+    pids.forEach(function(pid) {
+      var src = photos[pid];
+      var dst = grmState.dbState[pid] || (grmState.dbState[pid] = { flag: 'none', has_species_keyword: false });
+      // Only refresh has_species_keyword — don't clobber flag, since the user
+      // may have updated it in the modal since the initial seed and the diff
+      // logic compares against the opened-with snapshot.
+      dst.has_species_keyword = !!(src && src.has_species_keyword);
+    });
+    renderGroupModal();
+    grmUpdateApplyLabel();
+  }).catch(function() { /* network error: keep stale state, no UI thrash */ });
+}
+
 // Diff the current modal zones against the DB snapshot we opened with.
 // Returns counts of new flag/reject/clear writes plus tag adds that the
 // apply call will perform.
diff --git a/vireo/tests/test_pipeline_group_apply.py b/vireo/tests/test_pipeline_group_apply.py
index f4e7a35..e0a570e 100644
--- a/vireo/tests/test_pipeline_group_apply.py
+++ b/vireo/tests/test_pipeline_group_apply.py
@@ -151,6 +151,59 @@ def test_apply_records_edit_history_for_undo(app_and_db):
     assert 'keyword_add' in actions
 
 
+def test_apply_keyword_add_cancels_pending_keyword_remove(app_and_db):
+    """If a previous edit queued ``keyword_remove`` for the species and the
+    user re-tags the same photo via the burst modal, Apply must cancel the
+    pending removal — not leave both queued — so ``sync_to_xmp`` doesn't
+    strip the keyword in XMP after the add (it applies removals after adds
+    when both are present)."""
+    app, db = app_and_db
+    pid = _photo_ids(db)[0]
+    species = 'Coyote'
+
+    # Pre-state: a pending keyword_remove exists for the species (e.g.
+    # user removed the keyword in the regular review page earlier and
+    # hasn't synced to XMP yet). The DB tag itself is gone, so the
+    # apply endpoint will re-tag and route the queue write through
+    # _queue_keyword_add — which is what should cancel the pending
+    # remove instead of queueing a contradictory add.
+    db.queue_change(pid, 'keyword_remove', species)
+    pending_remove = db.conn.execute(
+        "SELECT COUNT(*) AS n FROM pending_changes "
+        "WHERE photo_id = ? AND change_type = 'keyword_remove' AND value = ?",
+        (pid, species),
+    ).fetchone()['n']
+    assert pending_remove == 1
+
+    client = app.test_client()
+    resp = client.post('/api/pipeline/group/apply', json={
+        'picks': [pid],
+        'rejects': [],
+        'candidates': [],
+        'species': species,
+    })
+    assert resp.status_code == 200
+
+    # The pending keyword_remove must be gone, AND no keyword_add should
+    # have been queued (the cancellation is a wash, not a swap).
+    pending_remove_after = db.conn.execute(
+        "SELECT COUNT(*) AS n FROM pending_changes "
+        "WHERE photo_id = ? AND change_type = 'keyword_remove' AND value = ?",
+        (pid, species),
+    ).fetchone()['n']
+    pending_add_after = db.conn.execute(
+        "SELECT COUNT(*) AS n FROM pending_changes "
+        "WHERE photo_id = ? AND change_type = 'keyword_add' AND value = ?",
+        (pid, species),
+    ).fetchone()['n']
+    assert pending_remove_after == 0
+    assert pending_add_after == 0
+
+    # And the keyword should be on the photo (Apply re-tagged it).
+    kws = {k['name'] for k in db.get_photo_keywords(pid)}
+    assert species in kws
+
+
 def test_apply_rejects_conflicting_zones(app_and_db):
     app, db = app_and_db
     pids = _photo_ids(db)
-- 
2.43.0
```

---

jss367:
🤖 Same git-proxy 403 — push from this routine is still blocked.

Patch from the prior comment was applied verbatim on top of `e526e9027f`; local commit `fe44f2df3c0bdfc919914d92e8d46a5855477a5d`. Diff stat matches the patch exactly: `vireo/app.py` +7/-1, `vireo/templates/pipeline_review.html` +60/-1, `vireo/tests/test_pipeline_group_apply.py` +53/-0 (3 files, +119/-2 total).

Verification on the local commit:
- Mandated suite + `vireo/tests/test_pipeline_group_apply.py`: **934 passed** in 104s (matches the prior attempt's 922 + 12 = 934, including the new `test_apply_keyword_add_cancels_pending_keyword_remove` regression test).
- `ruff check vireo/app.py vireo/tests/test_pipeline_group_apply.py`: **clean**.

`git push origin review-burst-apply-close` returned HTTP 403 from the receive-pack endpoint on every attempt, including after the standard 2/4/8/16s backoff cycle:

```
error: RPC failed; HTTP 403 curl 22 The requested URL returned error: 403
send-pack: unexpected disconnect while reading sideband packet
fatal: the remote end hung up unexpectedly
```

Environment is `CCR_TEST_GITPROXY=1` with `origin` forced through `127.0.0.1:33779/git/jss367/vireo`, so the block is at the sandbox proxy, not at GitHub. Manual intervention needed to land the commit (apply the `git am` patch from the parent comment, or cherry-pick `fe44f2d` if accessible).

---

jss367:
🤖 Re-attempt confirms the proxy block is persistent — push still HTTP 403 from `127.0.0.1:41107/git/jss367/vireo` on the initial attempt and all four backoff retries (2/4/8/16s). No new information vs the prior four comments on this same situation; the routine cannot land this from inside the sandbox.

Re-applied the patch from [comment 4413495900](https://github.com/jss367/vireo/pull/792#issuecomment-4413495900) verbatim on top of `e526e9027f`:
- Local commit: `cd2a93e` (same diff stat: `vireo/app.py` +7/-1, `vireo/templates/pipeline_review.html` +60/-1, `vireo/tests/test_pipeline_group_apply.py` +53/-0; +119/-2 total).
- Mandated suite + `test_pipeline_group_apply.py`: **934 passed** in 86s.
- `ruff check vireo/app.py vireo/tests/test_pipeline_group_apply.py`: clean.

Verbose curl confirms the 403 originates at the sandbox proxy's receive-pack, not GitHub:

```
Pushing to http://127.0.0.1:41107/git/jss367/vireo
POST git-receive-pack (3245 bytes)
error: RPC failed; HTTP 403
```

Suggest treating this as a routine-environment escalation rather than another retry — `git push` from this routine has been 403'd on every attempt across the last several invocations. To unblock the PR, please apply the `git am` patch from [comment 4413495900](https://github.com/jss367/vireo/pull/792#issuecomment-4413495900) directly on `review-burst-apply-close`.

---

jss367:
🤖 Same git-proxy 403 as the prior routine attempts on this PR — `git push origin review-burst-apply-close` returned HTTP 403 from the receive-pack endpoint on every retry across the standard 2/4/8/16s backoff cycle. Local commit `6ad4baf` exists on top of `292792b`; sandbox proxy blocks the push.

**What this addresses** (the one new P1 from Codex review on `292792b127`):

> **P1** [`vireo/templates/pipeline_review.html` line 2868] — Keep Apply disabled when group state fetch fails. When `/api/pipeline/group/state` errors, the catch path seeds with an empty DB snapshot and then marks the modal as seeded/enables Apply; previously flagged/rejected photos get treated as unflagged candidates, and Apply would clear those existing flags to `none` in the DB, silently undoing prior review decisions.

**Fix** (single file, 32+/2-): replace the `grmSeedFromDbState({}, …)` empty-snapshot fallback in the `.catch` handler with a new `grmHandleSeedFailure()` that:

- Renders the items so the user can still inspect the burst and close out, but does **not** populate `picks`/`rejects` and does **not** set `grmState.seeded = true`. `grmApply()` already early-returns when `!grmState.seeded`, so the apply endpoint cannot be hit.
- Disables the Apply button with an explicit label (`"Unavailable — reopen to retry"`) and tooltip — instead of leaving it stuck on the generic `"Loading…"` placeholder forever.
- Shows a toast so the user understands what happened.

Closing and reopening the modal naturally retries the `/group/state` fetch — no extra "retry" button or recovery surface needed.

**Tests:** 934 passed (full mandated suite per CLAUDE.md). No code path other than the burst modal's own seed-failure handler changes; the apply endpoint and its tests are untouched.

The other 6 review threads on this PR are already addressed in HEAD (`292792b`) from prior commits — they only show as unresolved because Codex hasn't re-reviewed since. Apply with `git am` against `review-burst-apply-close`:

```patch
From 6ad4baf565bbfb28d3a094c88e22f6a39ab46032 Mon Sep 17 00:00:00 2001
From: Claude <noreply@anthropic.com>
Date: Sun, 10 May 2026 21:42:36 +0000
Subject: [PATCH] Keep burst-modal Apply disabled when /group/state fetch fails

---
 vireo/templates/pipeline_review.html | 34 ++++++++++++++++++++++++++--
 1 file changed, 32 insertions(+), 2 deletions(-)

diff --git a/vireo/templates/pipeline_review.html b/vireo/templates/pipeline_review.html
index a3eb876..8b4259e 100644
--- a/vireo/templates/pipeline_review.html
+++ b/vireo/templates/pipeline_review.html
@@ -2864,11 +2864,41 @@ function openGroupReview(encIdx, burstIdx, selectPhotoId) {
     grmSeedFromDbState(data && data.photos ? data.photos : {}, items, selectPhotoId);
   }).catch(function() {
     if (grmState.sessionId !== sessionId) return;
-    // Network/error fallback: seed from heuristic only.
-    grmSeedFromDbState({}, items, selectPhotoId);
+    // /group/state failed (network drop, server 500, etc). We *cannot* fall
+    // back to an empty DB snapshot here — doing so would treat every
+    // already-flagged or already-rejected photo as a candidate, and a
+    // subsequent Apply & Close would clear those flags to 'none' in the DB,
+    // silently undoing prior review decisions. Leave grmState.seeded = false
+    // so grmApply() refuses to send anything, surface the failure to the
+    // user, and keep the Apply button disabled with an explicit label so
+    // there is no question about what's happening.
+    grmHandleSeedFailure();
   });
 }
 
+// Called when /api/pipeline/group/state fails on modal open. Keeps the modal
+// in a non-applicable state: Apply stays disabled (with a label users can
+// read instead of a generic "Loading…" stuck forever), and the items render
+// so the user can still inspect the photos and close out. Closing and
+// re-opening the modal retries the fetch.
+function grmHandleSeedFailure() {
+  // Render the items so the modal isn't an empty void — but do NOT populate
+  // picks/rejects (no DB snapshot to do it from) and do NOT mark seeded, so
+  // grmApply() early-returns if somehow invoked.
+  renderGroupModal();
+  var applyBtn = document.getElementById('grmApplyBtn');
+  if (applyBtn) {
+    applyBtn.disabled = true;
+    applyBtn.style.opacity = '0.5';
+    applyBtn.style.cursor = 'not-allowed';
+    applyBtn.textContent = 'Unavailable — reopen to retry';
+    applyBtn.title = 'Could not load DB state for this burst. Close and reopen the modal to retry.';
+  }
+  if (typeof showToast === 'function') {
+    showToast('Could not load review state for this burst. Close and reopen to retry.', 'error');
+  }
+}
+
 // Seed picks/rejects on modal open. Photos already flagged/rejected in the
 // DB land in those zones; the quality-score heuristic only fills in zones
 // for photos with no flag set, so re-opening a burst respects prior choices.
-- 
2.43.0
```

---

jss367:
🤖 No action taken — the P1 fix is already on the branch.

The comment's patch would land on top of `292792b`, but HEAD is now `af1b6f5` ("Keep Apply disabled when /group/state fetch fails"), pushed manually by the maintainer ~1 hour after the original routine push got 403'd. That commit addresses the same P1 via a functionally equivalent path:

| Behavior the Codex P1 requires | `af1b6f5` (live) |
| --- | --- |
| Don't seed from empty DB snapshot on `/group/state` failure | ✅ |
| Leave `grmState.seeded = false` so `grmApply()` early-returns | ✅ |
| Render cards so user can still inspect the burst | ✅ |
| Disable Apply button with explicit label instead of stuck "Loading…" | ✅ (`"Could not load — reopen to retry"`) |
| Tooltip explaining cause | ✅ |

Minor implementation deltas vs. the patch in the prior comment:

- `af1b6f5` **adds an extra guard the prior patch lacked**: `grmUpdateApplyLabel` early-returns when `!grmState.seeded` (lines 3109-3113), so a later species-input keystroke can't clobber the error label. That's a strictly stronger guarantee than the prior patch had.
- `af1b6f5` **does not include a toast**. The prior patch added one — a UX nicety, not required to fix the P1. Happy to add it in a follow-up if wanted, but didn't push it on top here to avoid stacking churn on a fix the maintainer already chose to ship a specific way.

Re-applying the prior patch would either conflict with `af1b6f5` or duplicate it, so stopping here. Let me know if you'd like the toast added as a separate small commit.

---

jss367:
🤖 CI fix attempted but could not reproduce the failure locally — manual intervention needed.

**What I checked on `af1b6f5`:**
- `ruff check vireo/ tests/` with ruff 0.15.12 (the latest version, same as CI's unpinned `pip install ruff`) → `All checks passed!` (218 files, exit 0).
- Repeated with `--no-cache` and on a fresh clone of the same SHA → same result.
- The only annotation on the failing lint job is the generic `Run ruff check: Process completed with exit code 1.` with no specific rule/file/line.
- The job duration was 8s and the `test` job on the same workflow run passed. The most recent commit (`af1b6f5`) only modified `vireo/templates/pipeline_review.html` (HTML/JS, not lint-checked).

**Likely cause:** transient CI flake (e.g. `pip install ruff` hiccup or setup-python wobble on Python 3.14) rather than a real lint violation in the diff.

**Suggested action:** re-run the failed `lint` job from the Actions tab. If it still fails after a re-run, please paste the actual ruff output here and I'll fix the specific violation — without it I can't identify a root cause to push a fix for.

---

PR #798, opened 2026-05-10
[codex] Harden batch delete endpoint

jss367:
## Problem

`/api/batch/delete` had two high-risk paths:

- It accepted `disk_permanent` retry requests with raw filesystem `paths`, then called `os.remove` on each path. A caller could ask the app process to delete files that were never resolved from Vireo's database.
- It passed client-supplied `photo_ids` directly into `db.delete_photos`, whose lookup is intentionally global for internal callers. That let the route delete photos outside the active workspace.

## Fix Plan

1. Keep direct permanent deletion by `photo_ids`, but only after the route verifies the IDs belong to the active workspace.
2. Replace raw retry paths with short-lived server-issued retry tokens. Failed Trash attempts now return `{path, retry_token}`; the UI sends `retry_tokens` back if the user confirms permanent deletion.
3. Preserve `Database.delete_photos` compatibility for internal callers by adding an opt-in `verify_workspace` parameter that the API route uses.
4. Cover the dangerous cases with focused regression tests.

## Changes

- Added one-time in-memory delete retry tokens with a 10-minute TTL.
- Rejected raw `paths` on permanent-delete retry requests.
- Validated batch delete photo IDs as integers and active-workspace members before deleting.
- Updated the shared delete modal to send retry tokens instead of raw paths.
- Added tests for cross-workspace delete rejection, raw-path rejection, and token-based retry.

## Validation

- `python -m ruff check vireo/app.py vireo/db.py vireo/tests/test_delete_api.py`
- `python -m pytest vireo/tests/test_delete_api.py -q`
- `python -m pytest tests/test_workspaces.py vireo/tests/test_app.py -q`

### DIFF
--- a/vireo/app.py
+++ b/vireo/app.py
@@ -631,15 +631,27 @@ def create_app(db_path, thumb_cache_dir=None, api_token=None):
     def _consume_delete_retry_tokens(tokens):
         if not isinstance(tokens, list) or not tokens:
             raise ValueError("retry_tokens required")
-        paths = []
         now = time.time()
+        # Validate every token before consuming any. Otherwise a request like
+        # [valid, invalid] would pop the valid entry, return 400, and leave
+        # the caller with no way to retry — and in the disk flow the DB row
+        # is already gone, so the file would be stranded.
+        seen = set()
+        validated = []
         for token in tokens:
             if not isinstance(token, str) or not token:
                 raise ValueError("retry_tokens must be a list of strings")
-            entry = app._delete_retry_tokens.pop(token, None)
+            if token in seen:
+                raise ValueError("delete retry token is invalid or expired")
+            seen.add(token)
+            entry = app._delete_retry_tokens.get(token)
             if not entry or entry.get("expires_at", 0) < now:
                 raise ValueError("delete retry token is invalid or expired")
-            paths.append(entry["path"])
+            validated.append((token, entry["path"]))
+        paths = []
+        for token, path in validated:
+            app._delete_retry_tokens.pop(token, None)
+            paths.append(path)
         return paths
 
     def _get_db():
--- a/vireo/app.py
+++ b/vireo/app.py
@@ -12,6 +12,7 @@ import math
 import os
 import queue
 import re
+import secrets
 import subprocess
 import sys
 import time
@@ -125,6 +126,7 @@ logging.getLogger("werkzeug").addFilter(_QuietRequestFilter())
 # SQLite versions. Sized below 999 to leave headroom for additional bound
 # parameters in joined statements.
 _SQL_PARAM_CHUNK = 900
+_DELETE_RETRY_TOKEN_TTL_SECONDS = 10 * 60
 
 
 def _chunked(seq, size=_SQL_PARAM_CHUNK):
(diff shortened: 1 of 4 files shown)


jss367:
🤖 Attempted to address the Codex P2 review (reissue retry token after a failed permanent-delete retry), but **could not push to `codex/harden-batch-delete`** — the sandbox proxy returned `HTTP 403` on every push attempt (4 retries with 2s/4s/8s/16s backoff). This session was provisioned with a different designated branch (`claude/sweet-feynman-Pxpod`), so it cannot write to the PR's head branch. Manual intervention needed.

The fix below was implemented locally and the full test suite passes (954 tests, including a new regression test covering the reissue behavior).

### The fix

When `mode == "disk_permanent"` is invoked with `retry_tokens`, `_consume_delete_retry_tokens` pops every token before `os.remove`. On a transient `OSError` (e.g. file momentarily locked), the response previously returned `{"path": p}` only — leaving the caller with no token to retry, even though the DB row was already gone from the earlier `disk` flow. Reissue a fresh token in that error path:

```diff
--- a/vireo/app.py
+++ b/vireo/app.py
@@ -2683,7 +2683,14 @@ def create_app(db_path, thumb_cache_dir=None, api_token=None):
                     trashed += 1
                 except OSError:
                     log.warning("Permanent delete failed for %s", p, exc_info=True)
-                    trash_failed.append({"path": p})
+                    # Reissue a token so a transient failure (e.g. the file is
+                    # momentarily locked) doesn't strand the file: the DB row
+                    # was already removed in the earlier ``disk`` flow, so the
+                    # only way back to it is through another retry token.
+                    trash_failed.append({
+                        "path": p,
+                        "retry_token": _issue_delete_retry_token(p),
+                    })
             return jsonify({
                 "ok": True, "deleted": 0, "trashed": trashed,
                 "trash_failed": trash_failed,
```

### Regression test

```python
def test_api_batch_delete_disk_permanent_retry_reissues_token_on_failure(
    app_and_db, tmp_path, monkeypatch
):
    """If os.remove fails during a permanent-delete retry, the response must
    include a fresh retry_token so the client can try again. The DB row was
    already removed in the earlier ``disk`` flow, so without a new token the
    file is stranded with no API path back to it."""
    import vireo.app as app_module

    app, db = app_and_db
    client = app.test_client()

    target = str(tmp_path / "locked.jpg")
    Image.new("RGB", (10, 10)).save(target)
    token = "old-retry-token"
    app._delete_retry_tokens[token] = {
        "path": target,
        "expires_at": time.time() + 60,
    }

    original_remove = os.remove

    def flaky_remove(path):
        if path == target:
            raise OSError("simulated transient failure")
        return original_remove(path)

    monkeypatch.setattr(app_module.os, "remove", flaky_remove)

    resp = client.post("/api/batch/delete", json={
        "mode": "disk_permanent",
        "retry_tokens": [token],
    })

    assert resp.status_code == 200
    data = resp.get_json()
    assert data["ok"] is True
    assert data["trashed"] == 0
    assert len(data["trash_failed"]) == 1
    failed = data["trash_failed"][0]
    assert failed["path"] == target
    new_token = failed.get("retry_token")
    assert new_token, "expected a fresh retry_token to be reissued on failure"
    assert token not in app._delete_retry_tokens
    assert new_token != token
    assert new_token in app._delete_retry_tokens
    assert app._delete_retry_tokens[new_token]["path"] == target
    assert os.path.exists(target)
```

Added next to `test_api_batch_delete_disk_permanent_retry_with_token` in `vireo/tests/test_delete_api.py`.

### Validation

```
pytest tests/test_workspaces.py vireo/tests/test_db.py vireo/tests/test_app.py \
       vireo/tests/test_photos_api.py vireo/tests/test_edits_api.py \
       vireo/tests/test_jobs_api.py vireo/tests/test_darktable_api.py \
       vireo/tests/test_config.py vireo/tests/test_delete_api.py
# 954 passed
```

---

PR #813, opened 2026-05-12
[codex] Support recursive workspace folder roots

jss367:
## Summary

This makes workspace folders behave like recursive roots while preserving the existing explicit folder-link model internally.

- Linking a workspace folder now also links known descendant folders so existing workspace-scoped photo queries keep working.
- Removing a workspace folder removes that folder and its descendants from the workspace.
- Workspace-facing APIs now return root folders only, so a library like `USA/2026` can appear as one folder instead of a long list of dated subfolders.
- Added regression coverage for recursive roots, historical descendant folders with missing `parent_id`, and root removal.

## Why

The `USA2026` workspace currently needs many dated subfolders linked directly because photos live in those child folders. This PR lets the user manage the parent folder as the visible workspace root without losing descendant photos from Vireo views.

New files still need to be discovered by a scan, but once descendant folders are known, the workspace can show the parent root instead of every child.

## Validation

- `python -m pytest tests/test_workspaces.py vireo/tests/test_scanner.py::test_scan_links_root_when_all_files_skipped vireo/tests/test_scanner.py::test_scan_links_restrict_dirs_when_all_files_skipped -q`
- `python -m pytest vireo/tests/test_db.py -k "folder_tree or workspace_extensions or get_photos_scoped or count_photos" -q`

### DIFF
--- a/.github/actions/fire-routine/action.yml
+++ b/.github/actions/fire-routine/action.yml
@@ -47,6 +47,11 @@ runs:
         echo
 
         if [[ "${status}" -lt 200 || "${status}" -ge 300 ]]; then
+          error_type=$(jq -r '.error.type // empty' < "${response}")
+          if [[ "${status}" == "429" && "${error_type}" == "rate_limit_error" ]]; then
+            echo "::warning::Routine fire skipped because the routine provider returned a rate-limit/quota error."
+            exit 0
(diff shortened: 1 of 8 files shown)


jss367:
@codex

PR #853, opened 2026-05-15
[codex] Preserve selected photo position in collections

jss367:
Keeps a single focused photo anchored when switching into a collection that contains it, including loading further collection pages until the photo is rendered. If the photo is not present, the selection still clears, and multi-select state continues to clear to avoid hidden batch actions. Adds e2e coverage for preserving the selected card and its grid offset. Also makes the PR Agent routine action treat provider-side 5xx responses as warnings so external routine outages do not fail CI. Validated with `python -m pytest tests/e2e/test_collection_context_menu.py -q`, `python -m pytest tests/e2e/test_browse_single_select.py -q`, `git diff --check`, and YAML parsing for `.github/actions/fire-routine/action.yml`.

### DIFF
--- a/.github/actions/fire-routine/action.yml
+++ b/.github/actions/fire-routine/action.yml
@@ -57,13 +57,13 @@ runs:
         echo
 
         if [[ "${status}" -lt 200 || "${status}" -ge 300 ]]; then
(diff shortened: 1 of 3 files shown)


jss367:
@codex

jss367:
@codex

PR #880, opened 2026-05-17
[codex] Defer lightbox 1:1 until high-res source loads

jss367:
## Summary

- Defer Browse lightbox 1:1 zoom until the source tier needed for that zoom has loaded.
- Preserve the click/key zoom anchor across the deferred snap.
- Show a `loading 1:1` badge while the high-resolution source is loading.
- Add E2E coverage for both unknown-size and known-size 1:1 requests so `/full` is not enlarged as a temporary soft 1:1 view.

## Root Cause

The previous device-pixel 1:1 fix corrected the final math, but the lightbox still applied the 1:1 transform immediately and only then swapped from `/full` to `/original`. For large photos or RAW-backed originals, that meant users could spend visible time looking at an upscaled preview and conclude the zoom was still soft.

## Validation

- `python -m pytest tests/e2e/test_browse_lightbox.py tests/e2e/test_lightbox_context_menu.py`
- `git diff --check`

## Summary by CodeRabbit

* **Tests**
  * Added e2e coverage verifying the lightbox waits for the high-resolution original before snapping to native 1:1; added a test that holds the original response to confirm deferred behavior and broadened keyboard-mode assertions to accept either immediate zoom or a pending 1:1 transition.

* **Bug Fixes**
  * Preserve the user’s click/center anchor during deferred 1:1 upgrades and clear stale anchors on open.
  * Show a “loading 1:1” indicator while high-res assets load.
  * Improve source-tier selection, swap and retry behavior for more reliable high-resolution handling.

[Review Change Stack](https://app.coderabbit.ai/change-stack/jss367/vireo/pull/880?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/tests/e2e/test_browse_lightbox.py
+++ b/tests/e2e/test_browse_lightbox.py
@@ -378,6 +378,117 @@ def test_browse_lightbox_one_to_one_nav_falls_back_when_original_fails(live_serv
     assert "/full" in page.locator("#lightboxImg").get_attribute("src")
 
 
+def test_browse_lightbox_restored_pending_one_to_one_waits_for_fallback_after_initial_original_fails(
+    live_server, page
+):
+    """A restored pending 1:1 must not snap on /full after initial /original fails."""
+    page.add_init_script(
+        "Object.defineProperty(window, 'devicePixelRatio', { value: 1, configurable: true });"
+    )
+    full_svg = (
+        '<svg xmlns="http://www.w3.org/2000/svg" width="600" height="400" '
+        'viewBox="0 0 600 400"><rect width="600" height="400" fill="#274"/></svg>'
+    )
+    fallback_svg = (
+        '<svg xmlns="http://www.w3.org/2000/svg" width="2560" height="1600" '
+        'viewBox="0 0 2560 1600"><rect width="2560" height="1600" fill="#642"/></svg>'
+    )
+    page.route(
+        "**/photos/*/full",
+        lambda route: route.fulfill(body=full_svg, content_type="image/svg+xml"),
+    )
+    page.route(
+        "**/api/photos/1",
+        lambda route: route.fulfill(
+            json={
+                "id": 1,
+                "filename": "restored-pending.jpg",
+                "width": 4000,
+                "height": 2500,
+                "flag": "none",
+                "wildlife_excluded": False,
+            }
+        ),
+    )
+    page.route("**/photos/*/original", lambda route: route.abort())
+    held_fallback = {}
+
+    def hold_fallback(route):
+        if "released" in held_fallback:
+            route.fulfill(body=fallback_svg, content_type="image/svg+xml")
+        else:
+            held_fallback["route"] = route
+
+    page.route("**/photos/*/preview?size=2560", hold_fallback)
+    page.route("**/photos/*/preview?size=3840", hold_fallback)
+
+    url = live_server["url"]
+    page.set_viewport_size({"width": 900, "height": 700})
+    page.goto(f"{url}/browse")
+    page.locator(".grid-card").first.wait_for(state="visible")
+
+    page.evaluate(
+        """() => {
+            const p = window.photos[0];
+            window._lbViewportByPhotoId[String(p.id)] = {
+                zoom: 1,
+                centerX: 0.5,
+                centerY: 0.5,
+                oneToOne: true,
+                pending1To1: true,
+            };
+            window.openLightbox(p.id, p.filename, window.photos);
+        }"""
+    )
+    expect(page.locator("#lightboxOverlay")).to_have_class("lightbox-overlay active")
+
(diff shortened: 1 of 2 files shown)


jss367:
Addressed the open review feedback in `804d43ba`:

- **Metadata-before-`/original` race (CodeRabbit):** `_lbApplyPendingOneToOneZoom()` now defers the snap until the displayed source tier is at least the tier 1:1 needs. I used a **tier-rank** comparison (`_lbSrcRank(current) < _lbSrcRank(required)`) rather than the suggested strict `requiredSource !== _lbCurrentSrcKey`. Strict inequality would regress 1:1-preserving arrow navigation (`preserveOneToOne` keeps `/original` current while `required` for a small photo is `full`, so `!==` would defer forever and never snap). Rank-based "current must be at least as sharp as required" fixes the soft-1:1 flash without that regression.
- **Stay soft on `/full` after `/original` fails (Codex):** the preload-error handler now keeps a pending 1:1 deferred and reschedules toward the sharpest remaining preview tier instead of snapping on an upscaled `/full` and rescheduling without a target.
- Strengthened `test_browse_lightbox_defers_one_to_one_until_original_size_known` to hold `/original` through the metadata response, assert it stays at fit while pending, then release and assert the snap lands on the high-res source.

The stored-anchor-preservation, second-toggle-cancel, and per-photo-anchor-reset threads were already resolved by earlier commits (`eebd8871`, `be0b07aa`, `010c02ee`) and re-verified against current code.

Standard suite: 963 passed; ruff clean. The e2e suite couldn't run in this environment (no Playwright browser) and is exercised by CI.

---

jss367:
Addressed the Codex review in `499fbdc3`.

**Fixed (still valid against current HEAD):**
- *Preserve deferred 1:1 intent across window resize* — the resize handler called `_lbSetZoom(_lbZoom, …)`, which cleared `_lbPending1To1` and (via its trailing `_lbScheduleSourceSwap`) retargeted the swap back to `/full`, so a resize during "loading 1:1" silently dropped the deferred snap. The handler now captures the pending intent/anchor, re-clamps, then restores it and re-arms the high-res swap. Added an e2e regression test (`test_browse_lightbox_resize_preserves_deferred_one_to_one`) that fails on the pre-fix code.

**Verified already addressed by earlier commits in this branch (skipped, no longer valid):**
- *Apply stored anchor when deferred 1:1 completes* → `_lbApplyPendingOneToOneZoom()` passes the saved anchor (commit `eebd8871`).
- *Second zoom toggle should cancel deferred 1:1* → `isZoomed` includes `_lbPending1To1` (commit `be0b07a`).
- *Reset deferred 1:1 anchor per-photo* → `_lbPending1To1Anchor` cleared on every open (commit `010c02ee`).
- *Keep deferred 1:1 targeting fallback tier after /original fails* → handled in the `preloader.onerror` path (commit `804d43ba`; thread is outdated).

Validation: full unit/API suite passes (963 passed); targeted e2e lightbox tests pass in isolation. Pre-existing e2e flakiness/`arrows_preserve_one_to_one_zoom` (relies on real fixture image files) is unrelated to this change.

---

jss367:
🤖 Addressed the latest Codex review (commit `2544d5bff9`).

**Thread 11** ("Avoid canceling fallback upgrade during deferred-resize recovery") was the only genuinely-unaddressed finding. Its code fix already landed in `4fdb22c8` ("Preserve fallback source during resize"), which preserves the already-queued `_lbDesiredSrcKey` instead of re-deriving the swap target from a `/full`-based `_lbNativeZoom`. I added a dedicated regression guard in `0260bc3c`: `test_browse_lightbox_resize_preserves_post_original_failure_fallback` — press `z` with dims unknown → `/original` aborts (error path queues the sharpest remaining tier) → resize while that tier is still held → the deferred 1:1 intent and higher-tier target must survive (no soft-1:1 on `/full`), then snap on the fallback tier once released. Verified it passes against `4fdb22c8` and fails pre-fix.

The other still-open, non-outdated Codex threads are already resolved in HEAD (verified against the code):
- "Apply stored anchor when deferred 1:1 completes" — `onSwapLoad` now routes through `_lbApplyPendingOneToOneZoom()`, which honors `_lbPending1To1Anchor`.
- "Let second zoom toggle cancel deferred 1:1" — `isZoomed = _lbZoom > 1.001 || _lbPending1To1`.
- "Reset deferred 1:1 anchor per-photo" — `_lbPending1To1Anchor = null` on every open.
- "Resolve pending 1:1 when original fallback needs no swap" — handled by the `_lbDesiredSrcKey === _lbCurrentSrcKey` inline-resolve guard.

Validation: required pytest suite **963 passed**; resize/fallback/deferred e2e subset **6 passed**. (The unrelated `arrows_preserve_one_to_one_zoom` e2e failure reproduces on the untouched branch — missing fixture image, environmental.)

---

PR #947, opened 2026-05-31
[codex] Fix batch EXIF location accept

jss367:
## Summary

Fixes the Browse location EXIF suggestion flow so accepting a reverse-geocoded location honors the active multi-selection instead of only updating the detail photo.

## Root Cause

Location keywords use a separate API/UI path from generic keyword assignment because they carry Google place metadata, parent-chain location keywords, replacement semantics, and GPS sidecar sync. The EXIF Accept button still posted only to `/api/photos/<id>/location`, so the existing batch keyword behavior did not apply. The Browse detail refresh also did not reload the active smart collection, leaving fixed photos visible in `GPS Without Location Keyword` until a manual refresh.

## Changes

- Added `/api/batch/location` for assigning one Google place location to multiple photos.
- Routed EXIF Accept and Google place selection through the active Browse selection when multiple photos are selected.
- Refreshed keyword/sidebar/sync state after location edits and reloaded the active smart collection so photos leave `GPS Without Location Keyword` immediately.
- Added API and browser regressions for multi-photo EXIF location accept.

## Validation

- `pytest vireo/tests/test_photos_api.py -q -k "location"`
- `pytest tests/e2e/test_location_section.py -q`

### DIFF
--- a/tests/e2e/test_location_section.py
+++ b/tests/e2e/test_location_section.py
@@ -45,6 +45,17 @@ def _seed_exif_photo(live_server, lat=40.785091, lng=-73.968285):
     return photo_id
 
 
+def _seed_exif_photos(live_server, photo_ids, lat=40.785091, lng=-73.968285):
+    """Set lat/lng on the given photos."""
+    db = live_server["db"]
+    with db.conn:
+        for photo_id in photo_ids:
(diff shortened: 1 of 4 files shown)


jss367:
@codex review

PR #948, opened 2026-05-31
Select all matching Browse photos

jss367:
Updates Browse so Cmd+A selects every photo matching the current workspace, folder, collection, and active filters instead of only thumbnails already loaded in the grid. Adds IDs-only endpoints for filtered photo queries and collections, then uses them from the Browse selection flow. Raises capture-time preview and job payload limits to support large selections like 6,000-photo timezone corrections. Validated with `pytest vireo/tests/test_capture_time_api.py vireo/tests/test_collections_api.py vireo/tests/test_photos_api.py`.

## Summary by CodeRabbit

* **New Features**
  * "Select All" now selects all matching photos across current browse filters and search results, not just visible ones.
  * Increased batch operation limits: capture-time correction workflows now support up to 50,000 photos per operation.
  * Enhanced filtering for collections and browse endpoints.

* **Tests**
  * Added test coverage for large-selection workflows and collection filtering.

### DIFF
--- a/vireo/app.py
+++ b/vireo/app.py
@@ -2626,6 +2626,34 @@ def create_app(db_path, thumb_cache_dir=None, api_token=None):
             }
         )
 
+    @app.route("/api/photos/ids")
+    def api_photo_ids():
+        """Return every photo ID matching the current Browse filters."""
+        db = _get_db()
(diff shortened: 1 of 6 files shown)


jss367:
@codex

PR #981, opened 2026-06-16
Refresh Needs Identification after keyword edits

jss367:
Refreshes the active Browse smart collection after keyword add/remove flows so photos leave Needs Identification immediately when an identifying keyword is added. Adds a Playwright regression that tags a photo from the Needs Identification collection and verifies the grid reloads without that photo. Root cause: keyword mutations refreshed sidebar counts but left the active collection grid using stale results. Validated with Browse E2E coverage and focused backend Needs Identification rule tests.


jss367:
@codex

PR #1017, opened 2026-06-22
Surface ExifTool dependency in onboarding, scans, and Settings

jss367:
ExifTool is the metadata backbone for every scan, but a missing binary previously degraded scans silently (no capture date, GPS, or camera info) with only a line in the log. This adds `exiftool_available()`/`exiftool_status()` helpers and a `/api/exiftool/status` endpoint, surfaces the check in the welcome flow (advisory, never blocks Get Started) and the Settings → System section alongside MegaDetector/darktable. It also appends a warning to the scan step summary so a degraded scan no longer reads as a clean success, satisfying the `CORE_PHILOSOPHY` "no black boxes" rule. Added 11 tests covering present/missing/unrunnable cases plus the endpoint; the full mandated suite passes (1183 passed, 1 skipped).

### DIFF
--- a/vireo/app.py
+++ b/vireo/app.py
@@ -10533,7 +10533,15 @@ def create_app(db_path, thumb_cache_dir=None, api_token=None):
             vireo_dir=vireo_dir,
             thumb_cache_dir=app.config["THUMB_CACHE_DIR"],
         )
-        return jsonify({"ok": True, "imported": len(paths)})
+        # Audit import calls scanner.scan just like the standalone scan/import
+        # paths above. Without ExifTool the newly imported photos still lose
+        # capture date, GPS, and camera info; the frontend renders any warning
+        # as a toast so the user isn't told the import "succeeded" silently.
+        response = {"ok": True, "imported": len(paths)}
+        metadata_warning = _scan_metadata_warning()
+        if metadata_warning:
+            response["warning"] = metadata_warning
+        return jsonify(response)
 
     # -- Scan status (kept, non-job) --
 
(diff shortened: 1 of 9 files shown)


jss367:
🤖 CI fix attempt for run [27935167535](https://github.com/jss367/vireo/actions/runs/27935167535):

**Linux**: ✅ passed (3482 tests with coverage)
**Windows**: ✅ passed
**macOS**: ❌ `worker 'gw1' crashed while running 'test_pipeline_loader_failure_marks_classify_rows_failed_not_skipped'` at 96–100% through the 3482-test suite.

The crash is a pytest-xdist **worker process death**, not an assertion failure. The test itself only `monkeypatch`es `classifier_mod.Classifier` to a function that raises `RuntimeError`, so no native code is invoked — there's no plausible code path in this PR (which is about surfacing ExifTool status) that could turn this deterministic-looking Python-only test into a worker crash. The test passes locally on Linux, and the same test in the same file passed on Linux + Windows in this same CI run.

This signature (deep-suite macOS worker crash with `-n 2`) is consistent with the macOS GHA runner running out of memory or hitting a native-lib quirk under sustained parallel load — not a regression from this branch.

**Action taken**: re-queued the failed macOS leg via `rerun_failed_jobs`. No code fix pushed, since none of the PR's changes touch the failing path and "fixing" a flaky transient with code changes would mask, not solve, it.

Note: the PR's current HEAD (`7c4e2814`) hasn't triggered a new CI run on its own — the failing run was on the prior commit `211fdf4b`. If the macOS re-run still crashes (or once you're satisfied the prior run was the issue), pushing any trivial commit (or letting the next real push happen) will get the latest HEAD CI'd.

---

PR #1021, opened 2026-06-23
Fix red main: add "move" to nav-ids coverage test

jss367:
## Problem

`main` is currently **red**: `vireo/tests/test_db.py::test_all_nav_ids_covers_every_page` fails, which blocks the required `test-os` checks on every open PR (including #1020).

## Cause

#1019 added the Move page — `/move` route, `move.html`, and `"move"` registered in `ALL_NAV_IDS` (`db.py`) — but did not add `"move"` to the **hardcoded `expected` set** inside `test_all_nav_ids_covers_every_page`. The test just compares those two sets, so they now diverge:

```
Extra items in the right set: 'move'
```

## Fix

Add `"move"` to the test's `expected` set. The page is real, so the test set was simply stale. One line.

Verified: `test_all_nav_ids_covers_every_page` passes.

### DIFF
--- a/vireo/tests/test_db.py
+++ b/vireo/tests/test_db.py
@@ -11822,7 +11822,7 @@ def test_all_nav_ids_covers_every_page():
     expected = {
         "pipeline", "jobs", "pipeline_review", "pipeline_rapid_review", "review", "cull",
         "misses", "highlights", "life_list", "browse", "map", "variants",
-        "dashboard", "audit", "compare",
+        "dashboard", "audit", "move", "compare",
         "zoom_test", "settings", "workspace", "lightroom", "shortcuts",
         "keywords", "duplicates", "logs",
     }


jss367:
🤖 CI diagnosis: the `test-os (macos-latest)` failure is a flaky pytest-xdist worker crash, **not** a real test failure caused by this PR.

From the macOS job log:

```
[gw1] [ 97%] PASSED vireo/tests/test_pipeline_job.py::test_onnx_load_failure_skips_sentinel_when_files_verify_ok
[gw1] node down: Not properly terminated
replacing crashed worker gw1
[gw1] [ 97%] FAILED vireo/tests/test_pipeline_job.py::test_onnx_load_failure_skips_sentinel_when_files_verify_ok
```

The worker emitted `PASSED` for the test, then died 1 ms later ("Not properly terminated"). pytest-xdist then marked the just-completed test as `FAILED`. Ubuntu, Windows, lint, and website all passed; the change in this PR is a one-line edit to a hardcoded test set in `test_db.py` and has no relationship to the ONNX pipeline tests.

Reran the failed macOS job rather than pushing a speculative code change. If it fails again, that would suggest a genuine resource issue (e.g. macOS runner OOM during the pipeline test suite) worth investigating separately — but it should not block this PR.

---

jss367:
🤖 CI fix attempted — manual intervention needed.

The `test-os (macos-latest)` leg failed again on the **same test** as attempt 1: `vireo/tests/test_pipeline_job.py::test_onnx_load_failure_skips_sentinel_when_files_verify_ok`, with the same pytest-xdist worker crash signature (`[gw1] node down: Not properly terminated`). Attempt 1's comment predicted that a recurrence would mean "a genuine resource issue worth investigating separately" — that's where we are now.

What I verified before stopping:

- **The PR's actual change is correct.** The 1-line addition of `"move"` to the `expected` set in `test_all_nav_ids_covers_every_page` makes that test pass.
- **The crashing test is unrelated to this PR's diff.** It's an ONNX/pipeline test added in #561; this PR only edits `vireo/tests/test_db.py`.
- **The full PR test suite is green locally on Linux** (1171 passed):
  ```
  python -m pytest tests/test_workspaces.py vireo/tests/test_db.py vireo/tests/test_app.py \
    vireo/tests/test_photos_api.py vireo/tests/test_edits_api.py vireo/tests/test_jobs_api.py \
    vireo/tests/test_darktable_api.py vireo/tests/test_config.py
  ```
- **Both ONNX load-failure tests pass locally** on Linux when run in isolation, so the assertions themselves are sound.
- The Ubuntu, Windows, lint, and website legs all passed. The macOS leg already runs with `-n 2` xdist workers (per `6e53662b` "Make macOS/Windows CI legs resilient to slower/smaller runners"), so the usual OOM mitigation is already in place and isn't enough here.

Why I didn't push a code change:

- Pushing a one-line `@pytest.mark.skipif(sys.platform == "darwin", ...)` (or `xfail`) on `test_onnx_load_failure_skips_sentinel_when_files_verify_ok` would unblock CI, but it would silently suppress a real macOS-specific worker crash on a test that explicitly guards against the "Repair never sticks" UX bug. That's an out-of-scope test-quality decision for a 1-line red-main hotfix, and exactly the "speculative change" the routine instructions say to avoid.
- Re-running the macOS leg again is unlikely to help — two consecutive identical failures on the same test, on a leg that's already been hardened against this exact category of crash, is no longer "flake noise."

Suggested next steps (any one unblocks the merge):

1. **Merge admin-override / bypass the `test` check** for this PR — it's an unrelated macOS flake, the diff is a literal test-set typo fix, and main is currently red because of the test this PR repairs.
2. **Quarantine the test on macOS** in a follow-up PR (e.g. `@pytest.mark.skipif(sys.platform == "darwin", reason="known xdist worker crash, tracked in #TBD")`) with a tracking issue for the real fix.
3. **Investigate the macOS-specific crash** locally. Useful starting points: rerun with `-p no:xdist` on the macOS leg to confirm whether the crash is xdist-specific or also reproduces single-process; bisect against the closely related sibling test `test_onnx_load_failure_writes_verify_failed_sentinel`, which differs only in what `model_verify.verify_model` returns and passed cleanly in the same run.

Stopping here to avoid pushing a half-fix that buries a real macOS signal.

---

PR #1030, opened 2026-06-23
Stabilize lightbox 1:1 metrics test

jss367:
## Summary
- Tightens the browse lightbox 1:1 e2e wait so metrics are sampled only after native zoom and fit scale are known.
- Prevents the test from racing a pending 1:1 transition where _lbNativeZoom can still be null.

## Validation
- pytest tests/e2e/test_browse_lightbox.py --browser chromium -q

### DIFF
--- a/tests/e2e/test_browse_lightbox.py
+++ b/tests/e2e/test_browse_lightbox.py
@@ -542,7 +542,14 @@ def test_browse_lightbox_one_to_one_uses_device_pixels_and_natural_layout(live_s
     # applies synchronously — but under CPU contention the 'z' keydown can be
     # processed slightly after page.keyboard.press resolves. Wait for the snap
     # to land before sampling layout so the metrics read can't race it.
-    page.wait_for_function("window._lbZoom > 1.001")
+    page.wait_for_function(
+        """() => (
+            window._lbZoom > 1.001 &&
+            window._lbNativeZoom > 1 &&
+            window._lbFitScale > 0 &&
+            !window._lbPending1To1
+        )"""
+    )
     metrics = page.evaluate(
         """() => {
             const t = document.getElementById('lightboxTransform');


jss367:
🤖 CI fix attempt: **rerunning the failed macOS job**, no code change pushed.

**Diagnosis**

- `test-os (macos-latest)` failed because a pytest-xdist worker (`gw1`) crashed mid-run while executing `vireo/tests/test_pipeline_job.py::test_pipeline_continues_when_first_model_fails`. No traceback was emitted — the worker died, not the assertions.
- `test-os (ubuntu-latest)` and `test-os (windows-latest)` both passed.
- This PR's diff (8 +/- 1 lines, only `tests/e2e/test_browse_lightbox.py`) does not touch the failing test or anything it imports, so the change cannot be the cause.
- The symptom matches the macOS/Windows runner memory-pressure pattern already addressed by #1018 (xdist worker cap) and #1022 (lightbox e2e hardening). The `-n 2` cap is already in effect; the heavier ML/pipeline tests can still spike past the macOS runner's headroom intermittently.

**Action**

Re-running the failed jobs only — no code change. If the rerun passes, this was a runner-side flake. If it fails again at the same place, the test deserves a closer look (likely a real macOS-Python-3.14 segfault in a native dependency loaded by the pipeline setup) and I'd need human input on how aggressively to harden vs. quarantine. Manual intervention may then be needed.

---

jss367:
🤖 CI failure on `test-os (macos-latest)` was a pytest-xdist worker crash (`[gw1] node down: Not properly terminated`) in `vireo/tests/test_pipeline_job.py::test_pipeline_single_model_still_aborts_on_failure`. The crashed test is unrelated to this PR's diff (one e2e lightbox wait condition), and the same job passed cleanly on `ubuntu-latest` and `windows-latest`. Re-running the failed macOS job — will follow up if it fails again with a real diagnosis.

---

PR #1068, opened 2026-07-01
Add "Rescan Folders" menu action (folder or whole workspace)

jss367:
## What & why

Adds a discoverable way to reconcile Vireo with what's on disk after photos are **added or deleted outside the app**. Motivated by: *"if I delete a bunch of photos outside Vireo, how do I ask it to rescan that folder?"*

Key insight this addresses: a scan only ever **adds/updates** rows — it never prunes originals deleted on disk (that's the separate *Missing Originals* flow, previously only reachable via a banner that re-checks every 30 min). So a plain "rescan" wouldn't catch deletions. This ties both halves — new/changed **and** deleted — behind one honest **Rescan** action.

## How to use

Open **Rescan folders…** from either "dropdown at the top":
- **Workspace switcher dropdown** — `⟳ Rescan folders…` (works in browser *and* desktop).
- **Native Tools menu** → *Rescan Folders…* (desktop; cross-platform Tauri).

Pick scope — **entire workspace** or **a specific folder**. Running it starts an incremental scan for new/changed photos, then checks the same scope for deleted originals and opens the *Missing Originals* review (scoped, with a "Showing only: <folder>" note) so you can remove the ghosts. If none are deleted, you get a toast instead.

## Changes

**Backend**
- `db.get_missing_photos(folder_id=…)` — scopes to a folder subtree via a recursive CTE on `parent_id`.
- `GET /api/photos/missing?folder_id=` — workspace-link validated (404 unknown / 400 invalid id).
- `POST /api/jobs/scan-workspace` — scans all reachable workspace roots, reports any offline roots it `skipped`, 400s when there's nothing to scan.

**Frontend / native**
- Rescan modal + scoped Missing Originals modal in `_navbar.html`.
- Workspace-dropdown action + native Tools menu item (`src-tauri/src/menu.rs`).

**Tests** — `vireo/tests/test_rescan_scope.py`: folder scoping incl. subtree, endpoint validation, scan-workspace queueing, empty-workspace 400.

## Verification
- Full Python suite: **1168 passed, 1 pre-existing skip**.
- Live endpoint checks + a **Playwright end-to-end run** of the modal flow (dropdown → modal → folder-scoped review shows only the right ghost; workspace scope unscoped) — no JS console errors.

## Notes
- The desktop **Tools menu** item appears after a Tauri rebuild; the **workspace-dropdown** entry works immediately in the browser. (`cargo check` couldn't run in this environment — it fails on a pre-existing missing bundled sidecar under `src-tauri/binaries/`, before reaching the 3-line menu change, which is pattern-identical to the adjacent `TOOLS_SCAN_LIBRARY` lines.)
- The folder picker lists workspace **root** folders; picking a root rescans its whole subtree. Arbitrary sub-folders remain available via the sidebar right-click "Rescan this Folder."

### DIFF
--- a/src-tauri/src/menu.rs
+++ b/src-tauri/src/menu.rs
@@ -66,6 +66,7 @@ pub mod ids {
 
     pub const TOOLS_RUN_PIPELINE: &str = "tools_run_pipeline";
     pub const TOOLS_SCAN_LIBRARY: &str = "tools_scan_library";
+    pub const TOOLS_RESCAN: &str = "tools_rescan";
     pub const TOOLS_FIND_DUPLICATES: &str = "tools_find_duplicates";
     pub const TOOLS_BUILD_PREVIEWS: &str = "tools_build_previews";
     pub const TOOLS_SYNC_METADATA: &str = "tools_sync_metadata";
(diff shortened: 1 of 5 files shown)


jss367:
@codex

PR #1088, opened 2026-07-03
Document shipped mask-based local adjustments architecture

jss367:
## What changed

This pull request now documents the mask-based local-adjustment architecture that actually shipped. It replaces the original 1,574-line proposal with a concise as-built note covering:

- the saved recipe and mask-snapshot contracts
- geometry-aware mask alignment and feathering
- per-pixel subject/background tone
- two-branch local detail rendering
- stale-mask updates, overlays, presets, and per-photo Copy Settings behavior
- explicit differences between the original proposal and the smaller failure-safe implementation
- two optional future enhancements, tracked as candidates rather than unfinished phases

The original proposal remains available in this pull request's commit history.

## Why

The feature was implemented by [Local adjustments: renderer and recipe schema (#1096)](https://github.com/jss367/vireo/pull/1096) and [Local adjustments: editor controls and mask workflow (#1100)](https://github.com/jss367/vireo/pull/1100). Those implementation reviews deliberately simplified several proposed mechanisms: one snapshot per recipe, synchronous per-target copying, aspect-mismatch refusal, and a cleanup grace window.

Merging the old proposal unchanged would describe contracts that do not exist. Closing it would leave four source and test references pointing to a missing document. The rewritten note resolves those references and makes the shipped code authoritative.

## Validation

- Confirmed the pull request differs from current `main` by exactly one documentation file.
- Checked the documented ranges, cleanup window, bulk-copy error handling, routes, and implementation references against current code.
- `git diff --check origin/main...HEAD`
- `pytest -q vireo/tests/test_local_masks.py vireo/tests/test_local_render.py` — 24 passed


jss367:
~~Pushed a fix~~ — retracted: the PR agent's latest commit (d8b86d92) already lands the same fix for the RAW byte-compare staleness point via `mask.source_digest` (hash over the source mask file + prompt + detector version, compared source-side). The current doc revision addresses all 12 review comments; no further changes from me.

jss367:
🤖 PR-agent triage — this Codex round re-raises 19 comments whose resolutions are already in the current commit (`40288c7`). Each comment's anchor line still sits near its original position, but the surrounding paragraphs have grown dedicated sections that respond to it. Rather than push another commit that would trigger yet another review round on already-covered ground, here is a per-comment pointer into the current text.

| # | Codex anchor | Concern | Addressed at |
|---|---|---|---|
| 1 | L18 | Define RAW snapshot decode space against edit renders | L20–42 — "materializes one snapshot variant per decode mode…generated from a `load_image` in that same decode mode as the edit render that will consume it" |
| 2 | L80 | Persist per-variant prompt metadata for cheap stale check | L622–649 — "Every digest input is a cheap DB/file-metadata signal, so the staleness check never runs MegaDetector or SAM" |
| 3 | L167 | Allow negative local detail deltas | L148–176 — "per-region they are signed deltas in −100..100 … clamp to detail's actual 0..100 input range happens after the branch resolves" |
| 4 | L336 | Invalidate caches when backfill rewrites working copies | L351–401 — "Both paths invalidate the photo's tracked render caches whenever they touch a working copy" (preview + thumbnail + external-handoff) |
| 5 | L416 | Skip embedded-fallback RAWs during snapshot creation | L442–466 — "Snapshot creation uses the same signal … no `preserve_highlights` entry is written when snapshot creation hits the fallback" |
| 6 | L654 | Avoid detector work on every stale-check | L622–649 — same section as #2; digest inputs are all cheap DB/file signals so recompute needs no detection |
| 7 | L734 | Handle newly available decode bases | L651–717 — "a decode basis the current source state now expects but which has no entry in `mask.decodes` is also treated as stale (a 'materializable' variant)" with the RAW-gains-companion and RAW-repaired examples spelled out |
| 8 | L741 | Protect new mask files from concurrent GC | L752–778 — "Publish is atomic against a concurrent GC sweep, not only against crashes" — per-`photo_id` publish lock + grace-period fallback |
| 9 | L984 | Clamp resolved local tone deltas | L985–994 — "Each field's resolved per-pixel amount is clamped to that field's validated global range … before the per-pixel op consumes it" (with the `contrast=-200` counterexample) |
| 10 | L1060 | Return per-target recipes after resnapshot | L1090–1102 — response is now "a map keyed by target `photo_id`" plus `pending` / `job_id` |
| 11 | L1060 | Move per-target SAM off the bulk request | L1066–1095 — "The resnapshot itself does not run inside the request … `JobRunner` background job" |
| 12 | L1074 | Persist resnapshot job before placeholder writes | L1104–1139 — durable `local_resnapshot_jobs` row committed in the same transaction as the placeholder writes; startup re-enqueues rather than mark-failed |
| 13 | L1078 | Clean up pending refs on cancel / terminal failure | L1307–1372 — "Cancel and terminal-failure paths run the same cleanup" (CAS-gated current-recipe strip + unconditional history/sync scrub) |
| 14 | L1078 | Block pending refs from export / handoff | L1216–1254 — "Rendered-output paths refuse to bake `pending:*` refs into permanent artifacts" (export skip-with-reason, Open External + iNat 409) |
| 15 | L1086 | Patch finalized refs in history + sync queues | L1161–1214 — finalization rewrites `edit_history_items.new_value` and pending XMP-sync rows in the same transaction as the ref swap |
| 16 | L1086 | CAS finalization against concurrent target rewrites | L1141–1159 — "Each placeholder ref is a compare-and-swap token" scoped per-`(job_id, target_id)` under the publish-lock |
| 17 | L1174 | Protect finalization vs in-flight XMP sync | L1192–1214 — "The XMP sync job skips `pending:*` refs so the queue row stays in place for finalization to overwrite" |
| 18 | L1493 | Persist consumed companion mask on RAW row | L1501–1525 — "The consumed `photo_masks` rows and `active_mask_variant` are copied onto the RAW primary in the same transaction that writes the resnapshotted recipe" |
| 19 | L1531 | Preserve companion masks during RAW pairing | L1488–1525 — "When the RAW primary has no usable mask of its own, pairing consumes the companion's mask before deleting the companion row" |

If any of these mappings looks wrong to you (a section that _looks_ like it addresses the concern but actually has a gap), reply on the specific row and I'll push a targeted fix. Otherwise these threads can be resolved — nothing further is queued for this round.

---

PR #1109, opened 2026-07-05
Fix standalone scan cancellation

jss367:
## Summary
Fixes standalone scan jobs so cancel requests are forwarded into scanner.scan through cancel_check. When scanner.scan raises its cancellation sentinel, the scan job now exits as cancelled, stops before later roots, and skips thumbnail generation instead of treating the cancel as a root failure.

## Validation
- pytest -q vireo/tests/test_multi_root_scan.py
- pytest -q vireo/tests/test_jobs_api.py::test_job_cancel_running_job_marks_cancelled vireo/tests/test_scanner.py::test_scan_cancel_check_aborts_before_discovery vireo/tests/test_scanner.py::test_scan_cancel_check_aborts_before_metadata_extraction

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved job cancellation during multi-root scans.
  * Scans now stop promptly when cancelled, rather than continuing through remaining roots.
  * Cancelled jobs now report clearer progress: the scan step is marked cancelled, thumbnail generation is skipped, and the job returns a cancelled result with the number of photos already indexed.

### DIFF
--- a/vireo/app.py
+++ b/vireo/app.py
@@ -12499,6 +12499,9 @@ def create_app(db_path, thumb_cache_dir=None, api_token=None):
                 runner.update_step(job["id"], "scan", current_file=message)
                 runner.push_event(job["id"], "progress", progress_payload)
 
+            def cancel_check():
+                return runner.is_cancelled(job["id"])
+
             vireo_dir = os.path.dirname(app.config["THUMB_CACHE_DIR"])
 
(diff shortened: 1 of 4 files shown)


jss367:
Addressed the review note in 1d9e4f69 by adding a shared ScanCancelled exception from scanner.py and using it in both standalone and pipeline scan cancellation handling.

PR #1117, opened 2026-07-05
Photo editor: dirty-navigation guard, crop-aware histogram, frozen fail state

jss367:
Fixes from an audit of the edit page (`photo_editor.html`). Complements #1116 (Enter-to-save); no overlapping hunks.

## What changed

1. **Unsaved edits were silently lost** on Back, navbar links, refresh, and tab close — only Prev/Next confirmed the discard. `goBackToBrowse()` now confirms like `editorNav`, and a `beforeunload` guard covers every other way off the page (with a suppress flag so Back doesn't double-prompt).
2. **Histogram/clip readouts now sample inside the active crop** — previously they sampled the whole uncropped preview, so a blown sky you'd already cropped away kept warning about highlight clipping. Auto Tone already sampled inside the crop ("the frame the user is keeping"); the Feedback panel now agrees. Crop-only edits (drag/aspect/full-frame) refresh the readouts via a debounced resample, mirroring the mask-overlay refresh pattern.
3. **Failed photo load left the editor live.** `loadPhoto`'s catch re-enabled the panels after resetting `recipe`/`savedRecipe` to `{}`, so a slider tweak + Save could replace the photo's real saved recipe with a from-scratch one (while the stage still displayed the *previous* photo's pixels). The editing surface now stays frozen on failure; Prev/Next/Back in the topbar remain usable.
4. **Mask overlay misaligned after window resize** — the resize handler repositioned the crop box but never the overlay.
5. **`pointercancel` never cleared a crop drag**, leaving the drag armed until the next pointerup.

## Testing

- Full pre-PR pytest suite: 1360 passed, 2 skipped.
- Verified in a real browser (Playwright against an isolated app instance on a DB copy):
  - Crop drag to 44%x41% → histogram canvas resampled after the 140 ms debounce, showing the cropped region's distribution.
  - Back with dirty recipe → `confirm("Discard unsaved edits to this photo?")` fired; dismiss stayed on the editor.
  - Closing a dirty page fired a `beforeunload` prompt; a clean page closed with none.
  - Mid-drag `pointercancel` → drag cleared; subsequent pointer moves left the crop untouched.
  - Subject local edit + Show Mask, then viewport resize 1600→1250 → overlay width tracked the image (900px→550px, aligned).
  - `/edit/999999999` → "Photo unavailable", panels frozen (`pointer-events: none`), Save stayed disabled even after a synthetic `setAdjustment()` call.

### DIFF
--- a/vireo/templates/photo_editor.html
+++ b/vireo/templates/photo_editor.html
@@ -642,6 +642,8 @@ var editorState = {
   navIds: [],
   loadSeq: 0,
   loading: false,
+  histogramTimer: null,
+  suppressUnloadPrompt: false,
 };
 
 function setEditorLoading(loading) {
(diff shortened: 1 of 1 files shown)


jss367:
Pushed a second commit fixing the remaining audit findings:

1. **True 1:1 zoom** — 100% now requests a native-resolution render (endpoint cap raised to 16384) and the status line reports the delivered dimensions, e.g. "(8288×5520 render)". Source selection no longer short-circuits to a 4096-capped working copy when the request needs more pixels: `working_copy_satisfies_recipe_render` is size-aware for uncropped recipes too (no behavior change for existing ≤3840 callers), and the empty-recipe sentinel in `/edit-preview` also fires when a working copy exists but is undersized for the request. Verified in-browser: 100% on an 8288×5520 NEF delivered exactly 8288×5520 (~10s RAW decode); Fit stays on the fast 1920 path.
2. **Aspect-ratio lock** — 3:2/4:3/1:1 buttons lock the ratio (shown active); corner drags preserve it exactly (verified: deliberately off-ratio drag landed at 1.5000); second click unlocks; Full Frame / Reset All / restore / photo change / 90° rotation clear it.
3. **Numeric crop entry** — x/y/w/h are now editable percent fields; w/h edits respect a locked ratio.
4. **Preset confirmations** — overwriting an existing preset name and deleting a preset both confirm (verified both dialogs fire; first save with a new name doesn't prompt).
5. **Straighten out-of-range recipes** — the slider widens beyond ±10° to fit a larger saved value instead of the browser silently clamping the thumb while the label shows the real number.
6. **No redundant renders** — dropped the cache-busting param and skip reloading when the identical render is already displayed; Before/After toggles on an unchanged recipe now issue zero requests (verified via request capture).

Tests: standard suite 1360 passed / 2 skipped, plus test_render_source.py, test_export.py, test_thumbnails.py, test_local_render.py (122 passed / 2 skipped). All items above re-verified with Playwright against a live instance, including a regression pass on the round-1 fixes (unlocked drags, anchor stability, pointercancel, Back guard, crop-aware histogram).

jss367:
Note: the second commit described in my comment above was pushed after this PR merged, so it never landed here — it's re-issued as #1118.

PR #1187, opened 2026-07-11
Extract photo color label routes into domain layers

jss367:
Extracts the existing photo color-label route group from the legacy application module into a focused Flask blueprint. Moves workflow coordination into a photo-label service and workspace-scoped SQL into a repository while retaining Database compatibility methods. Preserves all existing URLs, response shapes, status codes, workspace isolation, edit history, the route-contract snapshot, and /api/v1 behavior. Validated with 4,492 Python tests, Ruff, seven critical Playwright journeys, focused contract tests, and the enforced 100,000-photo performance benchmark.

### DIFF
--- a/vireo/app.py
+++ b/vireo/app.py
@@ -86,6 +86,7 @@ from render_source import (
 )
 from schema import ensure_schema
 from web.pages import pages_blueprint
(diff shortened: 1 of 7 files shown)


jss367:
@coderabbitai review

PR #1178, opened 2026-07-10
Normalize keyword punctuation variants

jss367:
Normalizes keyword comparison and creation so stray leading or trailing quote marks like `‘apapane` do not create separate keyword behavior from `apapane`.

The selection keyword suggestions and duplicate keyword listing now group by normalized name while preserving parent and type boundaries, and keyword cleanup can merge existing variants into the clean spelling.

Added focused regression coverage for new keyword creation, cleanup, duplicate reporting, and multi-select missing-keyword suggestions.

Validation: `pytest vireo/tests/test_app.py tests/test_workspaces.py`; `ruff check vireo/keyword_normalization.py vireo/app.py vireo/db.py vireo/tests/test_app.py tests/test_workspaces.py`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved keyword and species normalization so edge-quote/accent variants deduplicate to the canonical spelling, with safer handling of internal punctuation.
  * Rejects keyword inputs that become empty after normalization, preventing empty keywords from being created or queued.
  * Fixed duplicate/rename/merge and queued edit behavior to consistently keep the clean canonical variant while correctly retargeting tagging and undo/redo updates.
  * Updated XMP sync to reconcile by normalized equivalence, skip empty-normalized entries, and apply removals before additions.
* **Tests**
  * Added extensive regression coverage for normalization, deduping/merging, queue/history correctness, importer/scanner behavior, and XMP flat vs hierarchical removal.

PR #1215, opened 2026-07-12
Refactor keyword normalization to choke points + one-shot data repair

jss367:
Refactors the keyword punctuation normalization that landed in #1178 into a choke-point design, and adds the automatic one-shot data repair the merged version lacks. (#1178 was auto-merged while this was being built; this PR now supersedes its implementation on main rather than replacing an open PR.)

## Problem

Stray edge quotes in keyword names (e.g. `‘apapane` from hand-edited XMP) created duplicate keywords, mismatched pending sidecar changes, and curation rows that silently dropped out of highlight/life-list queries. #1178 fixed this by guarding each affected call site individually; 37 commits and 50+ review findings later the review loop still hadn't converged, because every fix added more surface for stored-variant edge cases. This PR reshapes the merged implementation: enforce the invariant in a few choke points and repair existing data once, so stored variants cannot exist at runtime and the per-call-site guards become unnecessary.

## Design

**Invariant: every stored keyword/species name is in `normalize_keyword_display()` form.**

- **`vireo/keyword_normalization.py`** (salvaged from #1178 verbatim): edge-quote stripping that preserves leading okina (U+02BB), internal acute accents, and non-ASCII case distinctions; `keyword_match_key()` uses an ASCII-only fold to stay in lockstep with SQLite's `COLLATE NOCASE`.
- **Write choke points** (`vireo/db.py`): `add_keyword`, `update_keyword`, and `_upsert_one_keyword` normalize names and reject empty-after-normalization input with `ValueError`; `queue_change` normalizes `keyword_add`/`keyword_remove` values so sidecar-bound spellings are always clean.
- **Rename/retype merge**: `update_keyword` detects a same-slot same-type peer under the effective (name, type) and merges into it instead of writing a duplicate (NULL parents bypass `UNIQUE(name, parent_id)`), returning the surviving id. Cross-type collisions under a real parent raise `ValueError` → documented 400. `_merge_keyword_into` rewrites pending sidecar changes from the source spelling to the survivor, scoped to tagged photos.
- **One-shot data repair** (`Database.normalize_keyword_data`, gated by the `keyword_names_normalized` db_meta marker — deliberately NOT `PRAGMA user_version`, since unmerged branch builds have already advanced the live DB past the next free version number, which would silently skip a version-gated migration): normalizes all existing keyword rows (merging normalized-equal duplicates with convergence over parent chains, folding `general` into specific-typed peers, dropping quote-only names and reparenting their children), rewrites pending change values, and moves species curation rows (`photo_preferences`, `species_representatives`, `species_highlights`) to the clean spellings. All-or-nothing; un-renameable cross-type collisions are logged.
- **Boundary layers** (salvaged from #1178 nearly verbatim — sidecar files on disk keep variants forever, so normalized matching is genuinely needed there): `xmp.py` removes keywords by normalized key with a flat-only mode, `sync.py` applies removals before additions and pairs normalization-only rename halves so Lightroom hierarchies survive, `scanner.py`/`importer.py` skip empty-normalized entries and compare by match key on ingest.
- **Routes** (`vireo/app.py`): catch the new `ValueError` → 400, and queue/record the *stored* spelling after `add_keyword` (which also applies the species casing convention). `_queue_keyword_add/_remove` normalize before cancellation matching. `PUT /api/keywords/<id>` returns the surviving `keyword_id` after a merge.
- **Two encounter-route fixes** surfaced by porting #1178's tests: the replacement path's previous-keyword lookup now accepts `type='taxonomy', is_species=0` rows (an explicit retype doesn't set the legacy column), and `previous_species` from the pipeline cache is normalized before lookup (cache files are not covered by the DB repair).

Net effect vs #1178: app.py went from +1,002 lines of per-route guards to ~+150; the "one more path that doesn't normalize" class of review finding is structurally impossible because the DB can't contain variants.

## Tests

Ported the regression suite from #1178 (the most valuable artifact of that review cycle): normalization unit tests, DB choke-point tests, rename/merge tests, and the full XMP/sync/scanner/importer boundary suites. Tests that asserted runtime tolerance of stored legacy variants were converted into repair tests (`vireo/tests/test_keyword_migration.py`, 11 tests covering rename-in-place, variant merging with tag union, general→taxonomy folding, cross-type separation, empty-name dropping with child reparenting, parent-chain convergence, pending/curation sweeps, okina preservation, and marker gating).

**Dry run against a copy of the real production DB**: the one-shot repair merged `‘apapane` (28 tagged photos) into the clean `apapane` and renamed `'anianiau` (90 tagged photos), leaving zero un-normalized names across keywords, pending changes, and curation tables, in 0.1s.

**Merge resolution note**: main's post-#1178 additions that existed only to tolerate stored variants (per-route guards, `sync_from_xmp` slot-dedup pruning) were dropped in favor of the repair; the source-side curation retargeting in `_merge_keyword_into` was kept — it's real cleanup behavior. The cross-slot homonym reconcile test from main was ported and passes against the simpler reconciler.

## Validation

- `python -m pytest tests/ vireo/tests/ -q` (full suite)
- `ruff check` clean on all changed files

Follow-up to #1178. Also carries #1207's detached-species-label fixes forward (re-applied during the merge).

## Summary by CodeRabbit

* **New Features**
  * Added automatic one-time standardization of legacy keyword spellings (including quote and case variants) using a canonical form.
  * Updated keyword-duplicate reporting to group variants consistently while respecting slot-specific identity (parent/type).

* **Bug Fixes**
  * Repeating “add keyword” on an already-tagged photo is now a true no-op (no extra pending changes or history).
  * Pending keyword add/remove actions now cancel reliably using stored canonical spellings, even when requests use variant spellings.
  * Species/taxonomy relabel/replace and highlights relabel now reject empty/quote-only inputs earlier and keep undo snapshots consistent with canonical spelling.
  * Keyword labeling sync now prunes differently to preserve distinct cross-slot homonyms.


jss367:
🤖 Re-verified the still-open Codex review threads against `3a685709cb`. This review submission is a banner-only recheck (no new inline comments), and every unresolved thread from earlier rounds is already addressed by commits on this branch:

| Thread anchor | Where it's addressed |
| --- | --- |
| `db.py:11793` retype merge species leak (thread 0 & 8) | `_merge_keyword_into` at `db.py:10016-10040` — `leaks_species_into_nontaxonomy` guard clears `is_species`/`taxon_id` on the survivor for `src.type=='taxonomy' OR src.is_species==1` into a non-taxonomy dst |
| `app.py:5944` no-op keyword add records edit (thread 1) | `already_tagged` short-circuit at `app.py:5936-5941` |
| `app.py:9153` relabel curation snapshot ordering (threads 3 & 27) | `db.resolve_species_display_name(species)` call at `app.py:9038-9053` runs before all `dst_existed` snapshots |
| `app.py:21000` species replacement uses `.lower()` (thread 5) | now uses `keyword_match_key(previous_species) != keyword_match_key(species)` at `app.py:20997-21000` |
| `db.py:9757` edit history retarget in migration (thread 7) | `_merge_keyword_into` edit-history retargets at `db.py:10134-10389` cover keyword_add/keyword_remove/prediction_accept/species_replace across `edit_history.new_value`, item new/old bare-string ids, and JSON `keyword_id`/`keyword_ids` payloads |
| `db.py:10257-10305` pre-existing survivor cleanup (threads 16, 20, 26, 29) | three targeted `DELETE`s at `db.py:10192-10277` (keyword_add/prediction_accept/species_replace new_value; keyword_remove old_value with "later add" guard; species_replace bare-string old_value) plus the JSON payload drop at `db.py:10331-10389` |
| `db.py:9864` normalize doesn't cancel opposite pending changes (thread 22) | opposite-type detection + paired delete at `db.py:9819-9850` mirrors `_queue_keyword_add`/`_queue_keyword_remove` runtime cancellation |
| `db.py:508` run one-shot repair before requests (thread 24) | explicit `init_db.normalize_keyword_data()` in `create_app` at `app.py:2696` (from commit `3a68570`) |

Local test suite: **1557 passed, 12 skipped, 1 failed**. The single failure is `test_failed_import_does_not_chain` in the routine's runtime environment — it relies on `chmod 0` to make a file unreadable, which is bypassed for UID 0, and reproduces identically on `main` (nothing PR-related). No new commit needed for this review.

---

PR #1214, opened 2026-07-12
Canonicalize species casing in highlights

jss367:
## Summary
Canonicalizes classifier species labels against existing keyword spelling so legacy rows like `Common Waxbill` merge into the `Common waxbill` Highlights bucket. The fix also canonicalizes Highlights API inputs and merges legacy highlight/representative curation keys so existing selections are not orphaned. Sentence-case handling now preserves mixed-case eponyms while normalizing all-caps labels.

## Validation
Ran focused Highlights, database casing, keyword normalization, and workspace merge regression tests; local commit hooks also passed.

PR #1275, opened 2026-07-16
Canonicalize species casing at prediction and curation boundaries

jss367:
Rebuilds closed #1214 on the #1215 choke-point foundation — 437 lines instead of ~930, no parallel read-tolerant API, no re-registered SQL function.

## Problem

Classifier labels are an external vocabulary with their own casing: label files say `Common Waxbill`, the keyword row says `Common waxbill`. Two user-visible bugs on main:

1. **Split buckets** — Highlights keyed unconfirmed photos by the raw prediction label and accepted photos by the keyword spelling, so one species showed as two buckets.
2. **Silently dropped stars** — starring a photo in a prediction-cased bucket wrote `species_highlights` keyed with classifier casing; the exact-match eligibility queries then never matched it again.

## Design

Same principle as #1215's XMP layers: canonicalize at the external-vocabulary boundary, keep stored state canonical, don't tolerate mixed state at read time.

- **Bucket collection** (`_collect_highlight_buckets`) keys predicted species through `db.resolve_species_display_name` (memoized per request) — the spelling `add_keyword` would store. Prediction rows stay verbatim (they're regenerated model output, like sidecar files).
- **Curation setters** (`set_photo_preference`, `add/promote/remove/move_species_highlight`, `clear_*`, species-filtered getters) and the two route body parsers canonicalize species input the same way, so bucket labels, eligibility prechecks, and stored rows agree on one string.
- **Eligibility fix**: `get_species_highlights`' fallback compares the photo's raw prediction label — now case-insensitively (`COLLATE NOCASE`), matching the dedupe rule used everywhere else. The exact compare marked every highlight starred from an unconfirmed bucket ineligible.
- **Eponym-safe casing** (carried from #1214): the `lower` convention no longer mangles `McKay's bunting` → `Mckay's bunting`, and ALL-CAPS label spellings sentence-case.
- **One-shot v2 repair** under its own `db_meta` marker re-runs just the curation case alignment, catching rows starred with prediction casing between #1215's repair and this fix. The pass is factored into `_align_curation_species_case`, shared with the v1 sweep.

## What was NOT carried from #1214

The six shadow `*_canonical` curation methods, the `vireo_normalize_keyword` UDF re-registration, and `add_prediction` write-canonicalization (which forced a reuse-rows restructure to avoid reclassify duplicates; unnecessary once display and accept both canonicalize).

## Tests

Two end-to-end regressions (bucket merging across casings; star → stored canonical → surfaces as highlighted → un-star under a third casing), curation setter canonicalization, eponym casing, and v2 marker gating (runs once, later mismatches untouched). Adapted from #1214's suite where applicable.

## Validation

- Focused suites: `test_app`, `test_db`, `test_keyword_migration`, `test_edits_api`, `test_photos_api`, `test_workspaces` — 1428 passed, ruff clean.
- Dry run against a copy of the production DB: v2 marker set, zero remaining curation/keyword casing mismatches.

Supersedes #1214.

## Summary by CodeRabbit

- **Bug Fixes**
  - Species spelling is now canonicalized end-to-end across highlights, photo preferences, and species curation, improving correct bucket merging when capitalization differs (including non-ASCII cases).
  - Species parsing/validation now fails fast when species is missing, and curation writes/read paths use the canonical stored spelling.
  - Added a one-time curation maintenance alignment that rekeys live curation and undo/redo snapshots to prevent case-mismatch recreation, while preserving intentionally distinct homonyms.
  - Improved sentence-case formatting for new casing conventions.

- **Tests**
  - Expanded regression coverage for casing mismatches, relabel/undo snapshot correctness, prediction-only scenarios, homonym ambiguity, and one-shot v2 alignment behavior.

### DIFF
--- a/vireo/app.py
+++ b/vireo/app.py
@@ -10390,6 +10390,28 @@ def create_app(db_path, thumb_cache_dir=None, api_token=None):
         # dst_existed=false; undo would then delete the user's
         # pre-existing highlight/preference/rep.
         species = db.resolve_species_display_name(species)
+        # Multi-homonym reconciliation. resolve_species_display_name
+        # preserves the caller's spelling when two intentionally-distinct
+        # root species keywords share a NOCASE key (legacy general
+        # ``Robin`` alongside taxonomy ``robin``) so bucket / parse /
+        # setter paths stay in agreement on one string. add_keyword's
+        # typed lookup below still prefers the taxonomy row, however, so
+        # a request submitted as ``Robin`` would snapshot dst_existed
+        # against ``Robin`` and then rename onto ``robin`` — leaving undo
+        # able to delete pre-existing ``robin`` curation rows it never
+        # created. Mirror add_keyword's ORDER BY here so snapshots and
+        # renames key on the exact spelling add_keyword will store. Rows
+        # with type NOT IN ('taxonomy', 'general') are excluded (matching
+        # add_keyword) so a deliberate individual/location/genre homonym
+        # doesn't misroute the target.
+        target_row = db.conn.execute(
+            "SELECT name FROM keywords WHERE name = ? COLLATE NOCASE "
+            "AND parent_id IS NULL AND type IN ('taxonomy', 'general') "
+            "ORDER BY (type = 'taxonomy') DESC, id ASC LIMIT 1",
+            (species,),
+        ).fetchone()
+        if target_row and target_row["name"]:
+            species = target_row["name"]
         error, status = _validate_highlight_photo_ids(db, photo_ids)
         if error:
             return json_error(error, status)
--- a/vireo/app.py
+++ b/vireo/app.py
@@ -542,7 +542,8 @@ def _photo_highlight_entries(db, photo_id):
         None, min_quality=0.0, photo_id=photo_id
     )
     buckets, _unidentified = _collect_highlight_buckets(
-        candidates, confidence_threshold=0.0
+        candidates, confidence_threshold=0.0,
+        canonicalize_species=_species_canonicalizer(db),
     )
     entries = []
     for bucket in buckets:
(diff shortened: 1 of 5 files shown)


jss367:
Force-pushed the branch back to `f97de1c8`, dropping the two most recent commits (`914fbb22`, `cbdd798f`).

Those addressed the Codex findings about species names starting with a **non-ASCII lowercase letter** (e.g. classifier `éclair Bird` → stored `Éclair bird`, where SQLite's ASCII-only `NOCASE` / `keyword_match_key` fold breaks). That input class can't occur in this catalog (bird/mammal common names are ASCII), and it's a non-terminating theme — every code path using `NOCASE` or `keyword_match_key` is a fresh finding for the same hypothetical species. `914fbb22` in particular added a per-row Python UDF (running its own keyword `SELECT`) inside the Highlights eligibility query, i.e. real cost on a hot path for an impossible input. Not worth carrying.

Everything through `f97de1c8` stands: the core canonicalization plus the genuine homonym fixes (cross-type homonyms and taxonomy-vs-hand-tagged-general, which do occur in real data). Removed the `claude-agent` label so the two non-ASCII threads aren't auto-reapplied. The related non-ASCII review threads can be resolved as won't-fix.

jss367:
Fresh Codex re-review at `f97de1c8` surfaced one new finding — flagging in a comment rather than auto-committing, since you just force-pushed back and explicitly removed the `claude-agent` label to stop the P2-drip cycle. Not re-adding the label.

**New thread**: [`vireo/db.py:10513`](https://github.com/jss367/vireo/pull/1275#discussion_r3596901065) — "Preserve relabel destination-existed flags during v2 history rewrite". Unlike the two open non-ASCII threads (12454, 12490) that fall under your won't-fix rationale, this one is ASCII-scope. Reasoned scenario:

1. Pre-fix relabel captured `species=<Prediction Casing>` in hl_prev with `dst_existed=False` for a photo P.
2. Later, a `common waxbill` (canonical) highlight row for P was created via a different path.
3. `_align_curation_species_case` renames `Common Waxbill` → `common waxbill` for P; `INSERT OR IGNORE` drops the pre-fix row and preserves the user's canonical row.
4. `_align_curation_history_species` rewrites the hl_prev species but leaves `dst_existed=False`. On undo, `_restore_relabel_curation` deletes the `common waxbill` row for P — which is now the user's pre-existing row, not the relabel-created one.

It's real in principle but requires the compounding conditions above (pre-fix relabel + independent same-photo canonical write + undo of that specific relabel). `_restore_relabel_curation` is already best-effort under stale state, and dropping the rewrite (or clearing `dst_existed` when a live-side coalesce is detected) both add code paths in a similar vein to the non-ASCII findings you dropped. Deferring to you on whether it's worth the surface.

---

PR #1276, opened 2026-07-16
Stabilize import and highlights E2E tests

jss367:
## Summary

Stabilizes two Chromium E2E regressions by synchronizing with application lifecycle state instead of fixed timing. The import test cancels its queued automatic refresh and waits for the stale duplicate preview to settle, while the highlights test waits for the lightbox visual handoff before sending the pick shortcut. This preserves the scenarios under test while matching the current debounce and transition safety behavior.

## Testing

pytest -q tests/e2e/test_import_page.py tests/e2e/test_highlights_initial_scope.py --browser chromium --tb=short (50 passed)


jss367:
@codex

PR #1349, opened 2026-07-22
website: update download links for v

jss367:
Automated update of download links following the v release. Auto-merges once required checks pass; deploy-website.yml runs on the resulting push to main.

### DIFF
--- a/website/src/pages/download.astro
+++ b/website/src/pages/download.astro
@@ -6,9 +6,9 @@ const base = import.meta.env.BASE_URL;
 const releaseUrl = (v: string) => `https://github.com/jss367/vireo/releases/download/v${v}`;
 
 // Per-platform versions — updated independently by CI when each platform builds
-const macosArm64Version = '0.28.0';
-const windowsVersion = '0.28.0';
-const linuxVersion = '0.28.0';
+const macosArm64Version = '';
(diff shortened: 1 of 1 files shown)


jss367:
🤖 Codex flagged a real P1 regression here, but this PR auto-merged (~30s after opening) before the review landed, so the fix agent has no branch to push to.

**The bug is live on `main` and deployed:** `website/src/pages/download.astro` sets `macosArm64Version`, `windowsVersion`, and `linuxVersion` to `''`, so all three primary download buttons on https://vireo… render as `https://github.com/jss367/vireo/releases/download/v/Vireo__aarch64.dmg` (etc.) — non-existent assets. Any visitor clicking Download right now gets a 404.

The routine's absolute rules forbid opening a new branch/PR, so I can't hotfix this from here. Manual intervention needed — likely either:

1. Revert commit `78c9250` on `main`, or
2. Fix the upstream CI job that generated this PR (whatever fills the version variable produced an empty string — the PR title `"website: update download links for v"` shows the same empty interpolation), and push a corrected `download.astro`.

Also worth investigating: this PR is configured to auto-merge on required-checks-pass, which means the Codex review cannot gate the merge. If Codex should be able to block releases like this, the required checks list needs to include it.

---

PR #1358, opened 2026-07-22
Stabilize asynchronous browser tests

jss367:
## Summary

Wait for Review collection options to be attached instead of requiring visibility, since option elements do not render independent boxes. Give the delayed free-text location save the full asynchronous test budget so resource-heavy E2E runs do not fail at Playwright’s default timeout. This stabilizes CI without changing application behavior.

## Validation

- Three originally reported tests passed together
- Filter-bar E2E suite: 18 passed
- Lightbox regression: 20 consecutive passes
- Ruff and git diff checks passed


jss367:
@coderabbitai review

PR #1389, opened 2026-07-28
Offer a darktable download from Settings

jss367:
## What

The "RAW Development (darktable)" panel in Settings was a dead end — it told you darktable was missing and left you to find it yourself. This makes it actionable: resolve the right build for your platform, download it with real byte progress, verify it, and hand off to the OS installer.

Design: `docs/superpowers/specs/2026-07-26-darktable-download-design.md`
Plan: `docs/superpowers/plans/2026-07-26-darktable-download.md`

## Key decisions

- **darktable only.** Adobe DNG Converter is proprietary and EULA-gated with no URL we can legitimately hotlink, so that row gets a "Get it from Adobe ↗" link and nothing more.
- **Latest release resolved from the GitHub API at click time**, not a pinned version — a pinned build goes stale between Vireo releases and a pinned URL 404s if an asset is renamed.
- **Integrity comes from the API's own SHA256 `digest`**, not a code-signature check. darktable does not successfully notarize its macOS builds ([upstream #19295](https://github.com/darktable-org/darktable/issues/19295), closed as not-planned; Homebrew disabled the cask over it) and its Windows installer is unsigned — a fail-closed `spctl`/Authenticode check would have deleted **every legitimate download**.
- **The Gatekeeper warning is conditional** on `com.apple.quarantine` actually being present. Measured: urllib downloads are not quarantined (LaunchServices applies that, not urllib), so it normally stays hidden rather than warning about a dialog you will never see.
- **Per-platform honesty.** macOS/Windows say "Download installer" because you finish the install; Linux says "Download and set up" because there we complete it.

## Bugs found and fixed along the way

- `find_darktable` never probed `/Applications/darktable.app`, so a normal macOS `.dmg` install reported "not found" — while `find_dng_converter` already probed its Adobe equivalent. Without this, a successful download still ended in ✗.
- **Asset matching is exact-suffix.** The release ships ~300 KB `.zsync` delta manifests whose names contain "AppImage"; a substring matcher would have "successfully installed" one of those instead of darktable.
- **`resolve_release` now returns `(release, reason)`.** A bare `None` conflated network failure with "no build for your platform", so a user behind a proxy or over GitHub's 60/hr rate limit was told their platform was unsupported — which is false.
- **`_url_is_trusted` had a path-traversal bypass** (`.../releases/download/../../../../attacker/x.dmg`), and `asset["name"]` was `os.path.join`ed unchecked so `../evil-x86_64.AppImage` could write outside the tools directory.
- **`APPIMAGE_EXTRACT_AND_RUN` was unconditional.** That flag makes the runtime skip the FUSE probe and *always* unpack the full ~178 MB squashfs, and the develop job loops per photo. Now gated on an actual AppImage, detected by magic bytes rather than filename (a renamed AppImage silently launched the GUI and hung for the full 120 s timeout).

## Testing

```
CLAUDE.md required suite:  1 failed, 1972 passed, 14 skipped
Suites touched here:       202 passed
ruff check vireo/ tests/:  All checks passed
```

The single failure is `test_api_exiftool_status_reports_missing`, confirmed failing on `origin/main` as well — pre-existing and unrelated.

Every task was mutation-verified. That caught three tests that passed under the exact bug they were written to catch, plus a `verify_digest` gap where a `blake2s` digest shaped identically to a SHA256 could only be rejected by algorithm name. The Settings panel was driven end-to-end in a real browser across the full completion matrix (success, no-digest, quarantined, Linux config-write, digest mismatch, cancelled, expired, connection loss).

## Known limitation

On Linux the AppImage still extracts once per photo during a develop job. The fix — `--appimage-extract` once at install time — is unverified because darktable's `AppRun` sets `CAMLIBS`, `IOLIBS`, `GIO_EXTRA_MODULES` and sources a GTK hook before exec'ing, so the extracted binary may not run without that environment. It needs testing on real Linux and is recorded as an open question in Task 7 of the plan rather than guessed at.

## Summary by CodeRabbit

* **New Features**
  * Added a “Get darktable” installation/download flow in Settings with availability pre-checks, streamed progress, resume/cancel handling, and platform-specific completion guidance (including macOS quarantine messaging).
  * Added API endpoints for darktable status diagnostics, install availability, and background download jobs with integrity verification.
* **Bug Fixes**
  * Hardened darktable detection and download safety (platform asset selection, path traversal protection, SHA256 verification), plus improved HTTP range resume correctness and robust cancellation.
* **Tests**
  * Expanded contract and unit/integration coverage for byte-accurate progress, cancellation, resumable/reuse behavior, and singleton job orchestration/asset-change handling.
* **Documentation**
  * Added new darktable download plan/specification documents.

### DIFF
--- a/docs/superpowers/plans/2026-07-26-darktable-download.md
+++ b/docs/superpowers/plans/2026-07-26-darktable-download.md
@@ -1057,6 +1057,43 @@ git commit -m "feat: verify downloaded darktable against the API's SHA256 digest
 - Modify: `vireo/darktable_install.py`
 - Test: `vireo/tests/test_darktable_install.py`
 
+### OPEN QUESTION — decide this before writing the Linux path
+
+Task 3's code review surfaced a real performance problem that this task is the
+right place to solve.
+
+**The problem.** Task 3 sets `APPIMAGE_EXTRACT_AND_RUN=1` when invoking an
+AppImage (gated on `_is_appimage`, so only AppImage users pay). But when that
+variable is set, the AppImage type-2 runtime does not probe FUSE first — it
+*unconditionally* unpacks the whole squashfs to a temp dir, runs, and deletes.
+darktable's AppImage is ~178 MB, and the develop job calls `develop_photo` once
+per photo in a plain loop (`app.py`). So a 200-photo develop job on Linux
+extracts and deletes 178 MB **two hundred times**, and that overhead counts
+against the 120 s per-photo timeout.
+
+**The proposed alternative:** run `./X.AppImage --appimage-extract` **once at
+install time** here in Task 7, and point `darktable_bin` at
+`squashfs-root/usr/bin/darktable-cli`. If it works it is strictly better — that
+layout is what darktable's walk-up-from-`argv[0]` resource lookup expects, it
+survives `realpath`, it needs no `argv[1]` prefix, no FUSE at any version, and
+costs nothing per export. The trade is ~500 MB of disk, once.
+
+**Why it is not already decided:** darktable's `AppRun` does real setup before
+exec'ing — it sets `CAMLIBS`, `IOLIBS`, `GIO_EXTRA_MODULES`, and sources
+`apprun-hooks/linuxdeploy-plugin-gtk.sh`. The extracted `darktable-cli` may not
+run correctly without that environment, depending on whether linuxdeploy patched
+RPATH. **This is an empirical question that must be tested on real Linux** — it
+cannot be settled by reading.
+
+**What to do:** test it on Linux before implementing the Linux branch. If the
+extracted binary works, take it and simplify (Task 3's `argv[1]` prefix and env
+var become dead code for our own installs, though keep them for user-configured
+AppImages). If it does not, keep the current approach and note the per-photo cost
+in the plan so it is a known limitation rather than a surprise.
+
+**Do not silently pick one.** If you cannot test on Linux, say so and leave the
+current approach in place with the cost documented.
+
 - [ ] **Step 1: Write the failing tests**
 
 Append to `vireo/tests/test_darktable_install.py`:
--- a/docs/superpowers/plans/2026-07-26-darktable-download.md
+++ b/docs/superpowers/plans/2026-07-26-darktable-download.md
@@ -1349,18 +1349,31 @@ def _clear_release_cache():
 
 Add `import pytest` to the file if it is not already there. Then:
 
+**CONTRACT CHANGE from Task 5's review.** `resolve_release()` returns a **2-tuple
+`(release_or_None, reason_or_None)`**, not a bare value. Bare `None` conflated
+four distinct outcomes — unsupported platform, no matching asset, asset rejected
+as suspicious, and network failure — so this route would have told a user behind
+a proxy or over GitHub's 60/hr rate limit *"No darktable build is published for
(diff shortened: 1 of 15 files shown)


jss367:
Checked the latest Codex review against `f78d76f9`. The review body carried only the standard "here are automated suggestions" preamble; the open (non-outdated, non-resolved) review comments Codex has posted on this PR all appear to already be addressed by that commit.

Verified against current code:

| # | Codex concern | Where it's handled |
|---|---|---|
| P1 | Bind POST to confirmed artifact (`app.py:18381`) | `expected_name/version/digest/size` read from body → mismatch returns 409 `darktable_asset_changed` (`app.py:18344-18426`) |
| P2 | Honor cancellation before hand-off (`app.py:18526`) | `runner.begin_uncancellable(job_id)` gate (`app.py:18523-18524`) |
| P2 | Reuse completed file after hash cancellation (`darktable_install.py:526`) | `already_complete` fast path (`darktable_install.py:498-517`) |
| P2 | Credit resumable partials in space preflight (`app.py:18474`) | `reusable_bytes` credits both completed dest and `.partial` (`app.py:18462-18474`) |
| P2 | Refresh cached release after 409 (`app.py:18417`) | `update_release_cache(asset)` before the 409 return (`app.py:18416`) |
| P2 | Ignore incomplete AppImages during detection (`develop.py:424`) | `os.access(..., os.X_OK)` gate — exec bit is the hand-off marker (`develop.py:424-425`) |
| P1 | Avoid extracting AppImage per photo (`develop.py:764`) | Extract-once cache: direct → extracted-AppRun reuse → per-photo only as last-resort fallback (`develop.py:719-801`) |
| P2 | Treat full partial as complete before resuming (`taxonomy.py:254`) | 416 handler promotes complete `.partial` to `dest` (`taxonomy.py:296-321`) |
| P1 | Reuse extracted AppImage on FUSE-less hosts (`develop.py:764`) | `_APPIMAGE_APPRUN_CACHE["extracted"]` reuses AppRun across `develop_photo` calls (`develop.py:741-762`) |
| P2 | Reserve space for AppImage extraction (`develop.py:780`) | `_has_space_for_extraction` requires ~3.5× source size before unpacking (`develop.py:96-108, 258-264`) |
| P2 | Warn about SmartScreen before installer launches (`settings.html:2428`) | SmartScreen/Gatekeeper text is included in the `confirm()` dialog **before** the POST (`settings.html:2290-2312`) |

The two remaining coderabbit comments on `docs/superpowers/plans/2026-07-26-darktable-download.md` (`_ALLOWED_HOSTS` snippet at L894, `verify_digest` split at L1095) are inside historical Task 5/Task 6 code blocks that show early iterative snapshots. The authoritative design spec (`docs/superpowers/specs/2026-07-26-darktable-download-design.md:55-97`) already documents the tightened final state ("`objects.githubusercontent.com` was removed from the allowlist after Task 5's…"), and the actual implementation in `vireo/darktable_install.py:48-53, 255-354` enforces github.com-only + normalized `/releases/download/` prefix and rejects non-SHA256 digests / non-hex values. Not editing the historical plan snippets for that reason.

Test suite (`CLAUDE.md`'s required list): 1989 passed, 16 skipped, 1 failure — `test_failed_import_does_not_chain` — which reproduces on `origin/main` (pre-existing, unrelated to this PR; missing exiftool in the sandbox). `ruff check vireo/ tests/`: clean. No code changes pushed.

---

PR #1392, opened 2026-07-30
Report unreadable sources in the mask stage instead of silent skips

jss367:
## What broke

Investigating why Process Review showed **666 of 984 photos as REJECT**, the cause turned out to be upstream of the review page entirely.

`extract_masks` treats a `render_proxy` returning `None` — meaning *the source file didn't open* — as an ordinary `skipped`, the same bucket as "SAM looked and found no subject here". The two mean opposite things, but both leave the photo unmasked, and `scoring.hard_reject_reasons` hard-rejects every unmasked photo with `no_subject_mask` (then again via the composite floor, since the subject features are missing).

Production run `pipeline-1785211429907-133`, Jul 28 06:03→18:31:

```
extract_masks: {masked: 395, skipped: 311, failed: 0, total: 706}
status: completed        errors: []
```

At 18:31:30 the SMB mount at `/Volumes/Photography` dropped — 310 `Input/output error` lines in ~16 seconds. Masks stop dead at photo id 464. The finalizer saw `em_failed == 0` and reported the stage completed, the job completed, and the user got auto-redirected into a review page where two-thirds of every encounter was a reject, with nothing anywhere explaining why.

The stage does have a source-offline guard, but it only runs as a pre-flight folder probe, so it cannot catch a mount that dies mid-loop.

## Changes

**`vireo/pipeline_job.py`**
- Count unreadable sources separately from benign skips; surface the count in the stage summary, the stage result dict, and the job error rollup.
- On a failed read, ask `_source_offline_reason` how far the problem reaches. A **folder-scoped** outage latches that folder so its remaining photos are accounted for without re-probing a dead server; a **mount-scoped** outage counts the untouched remainder and stops the stage rather than walking the rest of the worklist.
- Fail the stage when anything was unreadable. The pipeline page already maps a `[stage]` error to its card, so this alone produces an expanded Extract card, a banner naming the volume to reconnect, and no auto-redirect into the bogus review set. `regroup` still runs, so results are not lost.

**`vireo/templates/pipeline.html`**
- The Extract card read `ext.masks_created` / `ext.scored_count` — field names no backend has ever emitted — so its summary was blank on every run and the card went green "Done!" regardless of outcome. Now reads the real `masked` / `skipped` / `unreadable` / `failed` fields via a shared helper, and withholds the complete/"Done!" treatment when photos went unmasked.

## Before / after

Same scenario driven through the real stage code (share drops after 3 of 8 photos):

| | before | after |
|---|---|---|
| stage result | `3 masked, 5 skipped` | `3 masked, 0 skipped, 5 unreadable` |
| stage status | `completed` | `failed` |
| job errors | none | `[extract_masks] Fatal: 5 of 8 photos unreachable — … Reconnect the source and run Process again; until then these photos have no mask and Process Review rejects them as no_subject_mask.` |
| reads after outage | 5 | 1 |

## Tests

5 new tests in `vireo/tests/test_pipeline_job.py` (each watched fail first):
- mid-run outage fails the stage, accounts for every photo, stops re-probing
- mount-scoped outage stops the run instead of walking remaining folders
- a folder-scoped outage does **not** strand photos in still-reachable folders
- a single corrupt file in a healthy folder is reported, not silently skipped
- guard against over-correction: SAM finding no subject stays a benign `skipped` and leaves the stage `completed`

1 new browser test in `tests/e2e/test_pipeline.py` for the Extract card.

```
vireo/tests/test_pipeline_job.py, test_jobs_api.py, test_pipeline_api.py,
test_job_contract.py, test_local_processing.py, test_local_masks.py,
tests/e2e/test_pipeline.py        →  622 passed, 66 skipped

mandated suite (test_workspaces, test_db, test_app, test_photos_api,
test_edits_api, test_darktable_api, test_config, …)  →  1851 passed, 1 failed
```

The one failure is `test_app.py::test_api_exiftool_status_reports_missing`, which fails identically on a clean `main` checkout — pre-existing and unrelated.

## Known gap (not fixed here)

The standalone `/api/jobs/extract-masks` endpoint in `app.py` has the same silent-skip shape in its own copy of the loop. Fixing it properly means extracting a shared helper rather than duplicating the offline logic; happy to do that as a follow-up.

### DIFF
--- a/tests/e2e/test_pipeline.py
+++ b/tests/e2e/test_pipeline.py
@@ -706,3 +706,35 @@ def test_pipeline_extract_card_not_green_for_zero_mask_reasons(
     """)
     expect(page.locator("#numExtract")).not_to_have_class(re.compile("complete"))
     expect(page.locator("#statusExtract")).not_to_contain_text("Done!")
+
+
+def test_pipeline_card_shows_actionable_error_not_the_last_one(
+    live_server, page,
+):
+    """When a stage reports several errors, the card must keep the actionable
+    `Fatal:` one rather than whichever happened to be appended last.
+
+    The errors loop keyed by stage with last-value-wins, so a generic
+    "failed mask extraction" line appended after a source-outage Fatal buried
+    the reconnect instruction — the only message that tells the user what to
+    actually do (Codex #1392 P2).
+    """
+    url = live_server["url"]
+    page.goto(f"{url}/pipeline")
+    page.evaluate("""
+        _onPipelineComplete({
+          status: 'failed',
+          result: {
+            stages: {extract_masks: {
+              masked: 0, skipped: 0, failed: 1, unreadable: 1, total: 2,
+            }},
+            errors: [
+              '[extract_masks] Fatal: 1 of 2 photos unreachable — volume ' +
+              '/Volumes/Photography is not mounted. Reconnect the source ' +
+              'and run Process again.',
+              '[extract_masks] 1 of 2 photos failed mask extraction',
+            ],
+          },
+        });
+    """)
+    expect(page.locator("#statusExtract")).to_contain_text("Reconnect the source")
--- a/tests/e2e/test_pipeline.py
+++ b/tests/e2e/test_pipeline.py
@@ -809,3 +809,64 @@ def test_extract_step_outcome_clean_run_still_reads_done(live_server, page):
     assert outcome["clean"] is True
     assert outcome["status"] == "Done!"
     assert "5 masked" in outcome["summary"]
+
+
+def test_extract_step_outcome_surfaces_top_level_error_on_raised_worker(
+    live_server, page,
+):
+    """A worker that raises leaves `result: null` and stashes the exception
+    in the completion payload's top-level `errors`. Reading only nested
+    `result.errors` hid the real cause and rendered the generic
+    "Some photos have no mask" instead (Codex #1392 P2).
+    """
+    url = live_server["url"]
+    page.goto(f"{url}/pipeline")
+    outcome = page.evaluate(
+        "a => _extractStepOutcome(a[0], a[1], a[2])",
+        ["failed", None, ["SAM2 checkpoint corrupt: unexpected EOF"]],
+    )
+    assert outcome["clean"] is False
+    assert "SAM2 checkpoint corrupt" in outcome["status"], (
+        f"The top-level exception must reach the card when the worker "
+        f"raised; got {outcome!r}"
+    )
+
+
+def test_extract_step_outcome_labels_bare_cancellation(live_server, page):
+    """A cancel arriving before the worker returned a structured result must
+    read "Cancelled", not the generic "Some photos have no mask" fallback
+    (Codex #1392 P2).
+    """
+    url = live_server["url"]
+    page.goto(f"{url}/pipeline")
+    outcome = page.evaluate(
+        "a => _extractStepOutcome(a[0], a[1], a[2])",
+        ["cancelled", None, []],
+    )
+    assert outcome["clean"] is False
+    assert outcome["status"] == "Cancelled", (
+        f"A bare cancel must say so, not lie about missing masks; "
+        f"got {outcome!r}"
+    )
+
+
+def test_extract_step_outcome_prefers_nested_over_top_level_errors(
+    live_server, page,
+):
+    """When both nested and top-level errors exist, the nested (worker's own)
+    error is the actionable one — top-level is only the fallback for the
+    raised/cancelled paths.
+    """
+    url = live_server["url"]
+    page.goto(f"{url}/pipeline")
+    outcome = page.evaluate(
+        "a => _extractStepOutcome(a[0], a[1], a[2])",
+        ["failed", {
+            "masked": 1, "skipped": 0, "unreadable": 2, "failed": 0,
+            "total": 3,
+            "errors": ["Reconnect the source and run Extract again."],
+        }, ["duplicate of the nested error"]],
+    )
+    assert "Reconnect" in outcome["status"], (
+        f"Nested error wins over top-level; got {outcome!r}"
+    )
--- a/tests/e2e/test_pipeline.py
+++ b/tests/e2e/test_pipeline.py
@@ -608,3 +608,36 @@ def test_pipeline_shared_card_not_done_until_all_substages_complete(live_server,
         }});
     """)
     expect(page.locator("#pillClassify")).to_contain_text("Done")
+
+
+def test_pipeline_extract_card_reports_unreadable_photos(live_server, page):
+    """The Extract card must show what actually happened to the photos.
+
+    Two failures met here in production: the card read `masks_created` /
+    `scored_count`, field names the backend never emits, so its summary was
+    always blank; and unreadable sources were counted as ordinary skips. A
+    dropped share therefore rendered as an empty, green "Done!" card while
+    311 of 706 photos went unmasked — and every unmasked photo is then
+    hard-rejected in Process Review as `no_subject_mask`.
+    """
+    url = live_server["url"]
+    page.goto(f"{url}/pipeline")
+    page.evaluate("""
+        _onPipelineComplete({
+          status: 'failed',
+          result: {
+            stages: {extract_masks: {
+              masked: 395, skipped: 0, unreadable: 311, failed: 0, total: 706,
+            }},
+            errors: [
+              '[extract_masks] Fatal: 311 of 706 photos unreachable — ' +
+              'volume /Volumes/Photography is not mounted.',
+            ],
+          },
+        });
+    """)
+    summary = page.locator("#txtMasks")
+    expect(summary).to_contain_text("395 masked")
+    expect(summary).to_contain_text("311 unreadable")
+    expect(page.locator("#pillExtract")).to_contain_text("Failed")
+    expect(page.locator("#statusExtract")).to_contain_text("unreachable")
--- a/tests/e2e/test_pipeline.py
+++ b/tests/e2e/test_pipeline.py
@@ -641,3 +641,37 @@ def test_pipeline_extract_card_reports_unreadable_photos(live_server, page):
     expect(summary).to_contain_text("311 unreadable")
     expect(page.locator("#pillExtract")).to_contain_text("Failed")
     expect(page.locator("#statusExtract")).to_contain_text("unreachable")
+
+
+def test_pipeline_extract_card_does_not_claim_no_masks_were_needed(
+    live_server, page,
+):
+    """An empty worklist the backend emptied because something went wrong is
+    not the same as "nothing needed doing".
+
+    When mask extraction reports a `reason` (no detections, missing weights,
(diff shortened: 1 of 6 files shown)


jss367:
Round 4 addressed in `71b0c72d`.

**Fixed — benign early exits left a hollow job-tree row.** Round 3 sent the stage status straight to the runner, so a benign exit reported `skipped`, which `JobRunner.update_step` doesn't finalize. The helper now returns the runner step status separately: `failed` for an outage, `completed` for a benign skip, while the stage result keeps `skipped`.

**Fixed — already-masked photos counted as unreadable at preflight.** Added `_photos_with_active_mask` to exclude them. It deliberately does *not* re-check the stored SAM prompt against the current detection: a stale prompt means the mask is out of date, not absent, so the photo still isn't at risk of the `no_subject_mask` rejection this count exists to explain.

**Not changed — "Keep the source outage as the displayed rollup".** This was already fixed in 7aa4c71d, the commit under review. `_onPipelineComplete` no longer collapses with last-value-wins; it prefers a `Fatal:` error per stage (`pipeline.html` ~2877-2891), mirroring the backend's own headline pick. `test_pipeline_card_shows_actionable_error_not_the_last_one` pins it and fails against the old behavior.

jss367:
Round 6 addressed in `ccb87ac7`.

**Fixed** — standalone extract-masks job now returns `ok: false` with an actionable error when photos are unreadable, so the Jobs page, history, API clients and the completion event stop reporting success.

**Fixed** — the cancel branch reported the post-filter loop total next to a preflight-inclusive `unreadable`, persisting impossible counters like `2 unreadable, total: 1`. Test reproduces it exactly.

**Two findings are being re-reported against code that already fixes them:**

1. *"Keep the source outage as the displayed rollup"* — 4th time. `_onPipelineComplete` stopped collapsing errors last-wins in `7aa4c71d`; it now prefers a `Fatal:` error per stage (`pipeline.html` ~2877-2891). `test_pipeline_card_shows_actionable_error_not_the_last_one` fails against the old behavior.
2. *"Report unreadable files in the standalone Extract path"* — 2nd time, citing `app.py:27392-27396`. That line was changed in `dec3d39b` from `skipped += 1` to `unreadable += 1`, and both completion paths now share `_extractMasksClean`. `test_extract_masks_route_reports_unreadable_sources` pins it.

Happy to look again if there's a case these tests miss, but I don't want to churn the diff on findings that appear to be anchored to earlier revisions.

jss367:
🤖 The `address-review` routine fired for the round‑4 Codex review on `7aa4c71`, but two subsequent commits on this branch already addressed those items:

- `71b0c72` — early exit sends `completed` to the runner (matches r3687160559) and preflight excludes already-masked photos from `em_preflight_unreadable` (matches r3687160561).
- `dec3d39` — for a fully-cached offline folder, `at_risk_dropped_ids` (not `dropped_ids`) drives both the Fatal error and the failure latch (matches the follow-on r3687197437).

The remaining `is_outdated=false` thread on line 6376 ("Keep the source outage as the displayed rollup") was already resolved on the frontend in `7aa4c71` — `_onPipelineComplete` now prefers `[stage] Fatal:` entries over other rollups for the same stage, pinned by `test_pipeline_card_shows_actionable_error_not_the_last_one`. Backend ordering is untouched for that reason.

Round 5 (on `71b0c72`) and round 6 (on `dec3d39`) have their own newer feedback that will be picked up by their own routine fires. Nothing to push here.

---

jss367:
Round 10 — holding at `303076ab`; no changes pushed. Six findings, and the batch is self-contradictory:

| Finding | Status |
|---|---|
| Keep the source outage as the displayed rollup | 8th time — fixed in `7aa4c71d` |
| Include preflight drops in cancelled totals | 5th time — fixed in `ccb87ac7` |
| Attach latched outage details to failed early-exit steps | 3rd time — fixed in `0cb4a1bd` |
| Include rescued weak detections in preflight outcomes | 2nd time — fixed in `303076ab`, the commit under review |
| **Restrict weak preflight drops to contextual rescues** | new — and the direct inverse of the line above |
| Fail early exits whenever unreadable photos remain | new — see below |

The two weak-detection findings land on adjacent lines (343 and 344) and ask for opposite changes: one says the helper must reach *below* `detector_confidence` to catch rescued frames, the other says it must not count those frames without replaying the rescue's anchor-species test. `_preflight_mask_outcomes` can't replay that test — the photo is already out of `photos` — so the docstring states the trade-off explicitly and chooses over-reporting, because an over-reported outage is visible and correctable while an under-reported one is the silent completion this PR exists to fix.

On *"Fail early exits whenever unreadable photos remain"*: this is real but narrow, and I'm leaving it deliberately. It only fires when classify has already emitted its own source-offline Fatal, which is the pre-existing `already_flagged_classify` design — the job still fails, and classify owns the message. Making Extract independently fail the same outage would double-report it. Happy to change it if you'd rather Extract's row mirror the failure.

I'd merge here. Nine rounds produced nine real fixes; this round produced none that survive scrutiny.

jss367:
🤖 The `address-comment` routine fired on the round‑10 status comment above. That comment is the PR author explaining why nothing was pushed this round (four items re‑reported against code that already fixes them, two new items that contradict each other on adjacent lines, one intentional deferral to classify's own Fatal). It isn't a change request, so there is nothing for this routine to push.

Merging isn't in this routine's scope — the pr‑agent workflow squash‑merges on an approving review or a 👍 comment. If you'd like this to land, either action will trigger it.

---

jss367:
🤖 The `address-review` routine fired for the round‑9 Codex review on `0cb4a1bd`, but `303076ab` had already landed a fix for the one new item on that commit (r3687355839, "Include rescued weak detections in preflight outcomes").

I independently produced a fix and its regression test before pulling and seeing the existing commit — my version restructured `extract_masks_stage` to compute `contextual_weak_ids` on the pre-drop photo list and pass the exact rescue set into `_preflight_mask_outcomes`. The maintainer's `_preflight_candidacy_floor` approach is the simpler read of the same fix (any weak MDv6 detection counts as at-risk, deliberately erring toward over-reporting an outage), so I dropped my competing implementation rather than push a diff on top. Nothing to push.

The remaining `is_outdated=false` threads on this PR were addressed in earlier rounds:
- r3687118058 (line 6452, "displayed rollup") — fixed on the frontend in `7aa4c71d`, pinned by `test_pipeline_card_shows_actionable_error_not_the_last_one`.
- r3687235072 (line 6391, "include preflight drops in cancelled totals") — fixed in `ccb87ac7`.
- r3687322035 (line 6029, "attach latched outage details to failed early-exit steps") — fixed in `0cb4a1bd` via `_extract_masks_early_exit` returning `step_extra`, pinned by `test_extract_masks_early_exit_carries_outage_detail_to_the_step`.

---

jss367:
Round 14 addressed in `eee51242`.

**Fixed — top-level errors invisible on raised or bare-cancelled Extract jobs.** `_extractStepOutcome` only read `result.errors` (the nested worker result), so when `JobRunner` recorded `result: null` — the worker raised before returning — the completion event's actual failure message (stashed at the top level) never reached the card, and both raised jobs and cancelled-without-nested-errors jobs fell through to the generic "Some photos have no mask" (r3688507226). The helper now takes the top-level errors as a third arg and falls back to them when the nested list is empty; bare cancels get an honest "Cancelled" label. Three regression tests in `tests/e2e/test_pipeline.py`, each watched fail before the fix.

The other `is_outdated=false` threads on this PR trace to earlier rounds:

- r3687118058 ("displayed rollup") — fixed on the frontend in `7aa4c71d`, pinned by `test_pipeline_card_shows_actionable_error_not_the_last_one`.
- r3687235072 ("include preflight drops in cancelled totals") — fixed in `ccb87ac7`.
- r3687322035 ("attach latched outage details to failed early-exit steps") — fixed in `0cb4a1bd`.
- r3687403367 ("fail early exits whenever unreadable photos remain") — intentional deferral covered in the round-10 comment: this fires only when classify already emitted its own source-offline Fatal, and classify owns the message; making Extract independently fail the same outage would double-report it.
- r3687499188 ("carry unreadable counts into cancelled step warnings") — fixed in `0b3f91a2` (round 11).
- r3687575417 ("preserve every fatal outage for a stage") — from round 13, still open; will be picked up by its own routine fire.
- r3687465333 ("diagnose standalone decode failures before recommending reconnect") — still open; pre-existing item, not on the round-14 review.

---

jss367:
🤖 The `address-comment` routine fired on the round‑14 status comment above. That comment is the PR author summarizing what already landed in `eee51242` (the current head) and pointing each remaining `is_outdated=false` thread at the earlier commit that fixed it (or, for r3687575417, the round‑13 review that owns it, and for r3687465333, an out‑of‑scope pre‑existing item). It isn't a change request, so there is nothing for this routine to push.

One heads‑up unrelated to that comment: `mergeable_state` on this PR is currently `dirty` — the branch has merge conflicts with `main`. Merging isn't in this routine's scope, but if you'd like me to resolve the conflicts and rerun the suite, kick off a `resolve-conflicts` fire.

---

jss367:
Round 15 addressed in `4f4d8ff4`.

**Fixed — standalone extract-masks reported `error_count = N+1` for N real failures.** For every raised photo the loop appended `Photo <id>: mask extraction failed` to `job["errors"]`, then also returned an aggregate `N of Y photos failed mask extraction` in `result["errors"]`; `_run_job` folds `result.errors` back into `job["errors"]`, so every listener downstream (persisted history, completion event, API clients, Jobs page) overstated the failure count by one (r3688624550). Dropped the per-photo append and kept only the aggregate — per-photo tracebacks are still captured in the log via `exc_info=True`, the card still gets its actionable headline off `result.errors`, and `error_count` now matches the real count. Mirrors the pipeline stage's own pattern in `extract_masks_stage`. Pinned by `test_extract_masks_route_does_not_double_count_per_photo_failures` (2 == 1 before the fix).

The remaining `is_outdated=false` threads on this PR trace to earlier rounds and were addressed / deferred there:

- r3687118058 ("displayed rollup") — fixed on the frontend in `7aa4c71d`, pinned by `test_pipeline_card_shows_actionable_error_not_the_last_one`.
- r3687235072 ("include preflight drops in cancelled totals") — fixed in `ccb87ac7`.
- r3687322035 ("attach latched outage details to failed early-exit steps") — fixed in `0cb4a1bd`.
- r3687403367 ("fail early exits whenever unreadable photos remain") — intentional deferral (round 10): fires only when classify already emitted its own source-offline Fatal, and classify owns the message.
- r3687465333 ("diagnose standalone decode failures before recommending reconnect") — still open, pre-existing item, out of scope for this round.
- r3687499188 ("carry unreadable counts into cancelled step warnings") — fixed in `0b3f91a2`.
- r3687575417 ("preserve every fatal outage for a stage") — still open, from round 13; leaving as-is here to avoid stacking a UI-collapse change on top of a fresh backend fix.

Mandated suite: 2030 passed, 16 skipped. One pre-existing failure remains — `test_failed_import_does_not_chain` — which uses `chmod 0` to make a file unreadable and no-ops under root; it fails identically on a clean checkout with this change reverted.

---

PR #1395, opened 2026-07-31
Serve E2E browser tests from a threaded WSGI server

jss367:
## What broke

Two browser tests failed on a local full-suite run:

```
FAILED tests/e2e/test_compare_page.py::test_compare_treats_second_detected_species_as_additional
FAILED tests/e2e/test_sync_panel.py::test_rating_preview_responds_to_selected_sidecar_creator
       playwright._impl._errors.TimeoutError: Page.goto: Timeout 30000ms exceeded.
```

Neither reproduces deterministically — on this checkout `pytest -o addopts='' -q tests/e2e/`
is **644 passed, 1 skipped** both before and after this change, and each test passes alone.
So this PR does not claim to fix them outright. It removes a measured amplifier that turns
ordinary machine load into exactly those two failure shapes.

## The amplifier

`tests/e2e/conftest.py` and `tests/e2e/test_new_images_pipeline.py` both build their server
with `make_server("127.0.0.1", 0, app)`. `make_server` defaults to `threaded=False`, so the
whole browser suite has been running against a server that handles **one request at a time**.

Measured on a trivial app constructed identically:

| | request issued while a 3s handler is in flight |
|---|---|
| `threaded=False` (before) | 2.71s |
| `threaded=True` (after) | 0.00s |

The shipped app does not run this way. `main()` in `app.py`:

```python
# Waitress uses a fixed thread pool (default 4). Each open SSE stream
# (bottom-panel logs, job progress, import duplicate check) pins one
# thread for its whole lifetime, so the pool must be sized well above
# the plausible number of concurrent streams or page loads queue
# behind them and the app appears frozen.
waitress_serve(app, host="127.0.0.1", port=port, threads=16)
```

The browser suite was running against a pool of one — the exact condition that comment
exists to prevent.

## Why that produces these two failures

`page.goto` resolves on `load`, which waits for the page's thumbnails. On `/browse` the
navbar fires `checkNewImages`, `checkMissingFolders` and `pollJobs` at load, and
`/api/workspaces/active/new-images` blocks up to 0.5s waiting on a cold-cache walk. With one
server thread, the thumbnails queue behind those XHRs instead of overlapping. On a loaded
machine that stacks into the 30s default timeout — `test_sync_panel`'s failure verbatim. The
compare failure is the same shape one layer up: the assertion after `accept-subject` waits on
a `loadComparison()` refetch that has to get through the same single thread.

It is also a fidelity gap, not only a timing one: no browser test can open a real SSE stream
today, because one would pin the only thread forever. Every SSE-adjacent regression test in
the suite has to fake the stream in JS (`test_log_panel_perf.py` calls `addLpLine` directly)
rather than drive the endpoint.

## Safety

`_get_db()` already builds a `Database` per request via Flask `g` — the same isolation the
16-thread waitress server depends on in production. Werkzeug's `ThreadedWSGIServer` sets
`daemon_threads = True`, so `block_on_close` is `False` and the fixture's `server_close()`
still does not join, meaning teardown cannot hang on a leaked connection.

## Not the taxonomy path

Worth recording since #1393 diagnoses a nearby symptom: on this machine the taxonomy re-parse
is **not** what these two tests are hitting. `~/.vireo/taxonomy.json` has an atime of
2026-07-19 (atime updates are live on this APFS volume, verified), and no checkout contains a
`vireo/taxonomy.json`. `load_local_taxonomy()` therefore returns `None` instantly throughout
the browser suite. #1393 is still worth landing on its own merits — the per-request re-parse
is a real production cost, and pinning the taxonomy candidates in `conftest.py` closes a real
isolation hole — it just isn't the cause of these two failures here.

## Tests

- `pytest -o addopts='' -q tests/e2e/` — **644 passed, 1 skipped** (10m43s), unchanged from
  644 passed, 1 skipped (10m15s) on the same checkout before the change.
- `ruff check tests/e2e/conftest.py tests/e2e/test_new_images_pipeline.py` — clean.


jss367:
Addressed the P2 in `034e29a7`.

The finding is real and I confirmed the mechanism: `BaseServer.shutdown()` finishes the handler it is currently running before returning, so the single-threaded server guaranteed nothing was in flight by the time the fixture reached `_cleanup_app_resources()` and `db.close()`. `ThreadingMixIn` drops that — `server_close()` skips its join because werkzeug sets `daemon_threads = True`, which makes `block_on_close` False. And `test_folder_tree_reveal_fires_endpoint` really does return while its handler can still be inside `subprocess.run(["open", "-R", ...], timeout=5)`.

Measured directly: with the pre-drain teardown the handler had completed before teardown returned in 0 of 1 runs; with the drain, 1 of 1.

Note there were two fixes for this in flight — the agent pushed `InFlightMiddleware` in `5599286b` while I was writing `threaded_server.py` locally. I kept the thread-tracking one and removed the middleware, because:

- Tracking the handler thread is a superset of tracking the WSGI callable. The middleware released its counter on the response iterable's `close()`, but the thread is still alive past that — flushing to the socket, running `shutdown_request`.
- Its `drain()` returned False on timeout and both call sites discarded the result, so a wedged handler drained silently. This one warns with the thread names.
- It added `sys.path.insert` + `# noqa: E402` to two files; `tests/e2e/__init__.py` already exists and `from e2e.leaflet_stub import ...` is the established sibling import, so no path hack is needed.
- The middleware commit had no tests. `tests/e2e/test_threaded_server.py` now covers three properties without a browser: concurrency is preserved, teardown does not return while a handler runs, and a wedged handler is given up on within the bound.

I left `5599286b` in history rather than force-pushing over it; the PR squash-merges, so it collapses away.

**Tests:** `pytest -o addopts='' -q tests/e2e/` — 647 passed, 1 skipped (11m17s), 644 before the three new cases. No straggler warnings across the suite, so nothing currently leaks a handler past the 10s bound. `ruff check tests/e2e/` clean.

jss367:
Both P2s from the `034e29a7` review are addressed in `408ca861`. I checked each against CPython's source before changing anything.

**Register handler threads before starting them** — confirmed, and the reasoning is exact. `Thread.start()` waits on `_started`, and `_bootstrap_inner` calls `_started.set()` *before* `run()`, so `start()` can return with the handler body not yet begun and nothing registered.

Counting now happens in `process_request`, on the accept loop, before `ThreadingMixIn` creates the thread at all. The thread set is kept only to name stragglers in the warning; the counter is what `drain()` waits on.

Measured against the previous design, with handler start delayed past `serve_forever`'s 0.5s poll interval: in-flight count at accept was **0** where it should be 1, and teardown returned before the handler had run. Both hold now. Worth noting the delay has to clear that poll interval to be visible at all — at 0.3s the old design still passes, at 1.5s it does not — so the test fixture uses 1.5s and documents why. Two new cases cover it: the count at accept, and `stop_server` waiting for a handler that has not started.

**Remove the load-sensitive concurrency deadline** — agreed, and it was the sharper version of the problem: that assertion would have failed for exactly the machine-load condition this PR exists to tolerate. Replaced with the handshake you suggested — `/blocker` holds a thread until `/fast` has been served, so `/fast` completing at all proves concurrency, and a single-threaded server deadlocks rather than merely running slow. No wall-clock threshold left.

**Track handler threads before entering the WSGI app** — stale: it reviews `tests/e2e/_e2e_server.py`, which this branch deletes in `034e29a7`. It restates the same race as the first comment, which is fixed.

**Tests:** `pytest -o addopts='' -q tests/e2e/` — 649 passed, 1 skipped (9m44s); 644 before the five `test_threaded_server.py` cases. No straggler warnings anywhere in the suite. `ruff check tests/e2e/` clean.

PR #1398, opened 2026-07-31
Stop duplicate-only remote imports from rescanning the destination folder

jss367:
## What's wrong

`_run_remote_import_job`'s per-batch catalog scan (`vireo/import_job.py:1506`) ran without `incremental=True`:

```python
if landed or dup_skipped:              # local path uses `if landed:`
    scan_files = landed_paths | set(adopted_paths.keys())
    scan(destination, db,
         restrict_dirs=[dest_folder],
         restrict_files=(scan_files or None),   # → None when nothing landed
         ...)                                   # no incremental= → defaults False
```

For a **duplicate-only batch** — every card file already in the catalog — `landed` is empty, so `restrict_files` collapses to `None`. `scanner.scan` reads that as *no file filter* (`scanner.py:1621`, `1682`), so the call walks the **entire destination folder**, and with `incremental` defaulting to `False` the skip logic at `scanner.py:1998-2035` never runs. Every already-cataloged file in that folder gets re-read and re-hashed.

Re-importing a card into the same date folder it originally landed in means that folder already holds every twin, so this re-reads the whole folder.

## Impact

Found while diagnosing a real import that had been running 2h10m with zero files copied. It was re-reading EXIF from 984 already-imported photos in a single archive folder, over an SMB-over-Tailscale mount, at **~3 files/min** — sampled live:

```
04:02:49   DSC_6774.NEF … DSC_6798.NEF   (25-file batch)
04:03:29   DSC_6774.NEF … DSC_6798.NEF   (same batch, still going)
04:04:09   DSC_6774.NEF … DSC_6785.NEF   (split to 12 after timeout)
```

exiftool batches kept exceeding their timeout on the high-latency mount and halving (50 → 25 → 12 → 6 → 3), burning the split budget. Projected ~5.5h for one folder, producing nothing — every value it fetched was already in the catalog.

## The fix

Pass `incremental=True` at that call site. Fresh landings are unaffected: they have no catalog row yet, so the skip path can't fire for them.

The local path never had this bug — its guard is `if landed:`, so a duplicate-only batch skips this scan entirely and falls through to the dedicated dup-link scan, which already passes `incremental=True`. The other three `scan()` call sites in the file were already correct.

## Why the existing test missed it

`test_remote_import_dup_link_scan_does_not_reread_unchanged_twins` already asserted this, but parks its twins in `mount_base/unsorted` — a *different* folder from the batch destination `mount_base/2026/2026-07-03`. So `restrict_dirs=[dest_folder]` walked an empty directory and read nothing either way. Right assertion, wrong geometry.

The new test puts the twins in the destination folder. On the old code:

```
AssertionError: remote duplicate-only import re-read already-cataloged,
unchanged twins in the destination folder: [.../DSC_0001.jpg, .../DSC_0002.jpg]
```

It seeds `_mark_exif_extracted(db)` like its siblings, since scanner's incremental skip requires non-NULL `exif_data` — without that the test would pass locally and fail in CI where ExifTool isn't installed.

## Tests

| Suite | Result |
|---|---|
| New test, before fix | **fails** (re-read both twins) |
| New test, after fix | passes |
| `vireo/tests/test_import_job.py` | 105 passed, 1 skipped |
| CLAUDE.md required set | 2029 passed, 1 failed |
| `test_scanner.py` + `test_import_dedup.py` | 157 passed |

The single failure is `test_app.py::test_api_exiftool_status_reports_missing`, a pre-existing environmental failure on the dev machine (ExifTool *is* installed there, so the test's "missing binary" premise doesn't hold). Unrelated to this change.

### DIFF
--- a/vireo/import_job.py
+++ b/vireo/import_job.py
@@ -1516,10 +1516,22 @@ def _run_remote_import_job(job, runner, db, workspace_id, params):
             # files never become photo rows. See PR #1113 review.
             scan_files = landed_paths | set(adopted_paths.keys())
             try:
+                # ``incremental=True`` is what keeps a duplicate-only batch
+                # affordable. With no fresh landings and no adopted paths,
+                # ``restrict_files`` is None and this call walks the whole
+                # ``dest_folder`` — and re-importing a card into the same
+                # date folder it originally landed in means that folder
+                # already holds every twin. Non-incrementally that re-read
+                # and re-hashed all of them, turning a zero-copy import
+                # into an hours-long rescan on a network archive. Fresh
+                # landings are unaffected: they have no catalog row yet, so
+                # the skip path can't fire for them. Mirrors the dup-link
+                # scan below and the local path's equivalent.
                 scan(
                     destination, db,
                     restrict_dirs=[dest_folder],
                     restrict_files=(scan_files or None),
+                    incremental=True,
                     vireo_dir=params.vireo_dir,
                     thumb_cache_dir=params.thumb_cache_dir,
                     skip_working_copies=True,
--- a/vireo/import_job.py
+++ b/vireo/import_job.py
@@ -1516,22 +1516,32 @@ def _run_remote_import_job(job, runner, db, workspace_id, params):
             # files never become photo rows. See PR #1113 review.
             scan_files = landed_paths | set(adopted_paths.keys())
             try:
-                # ``incremental=True`` is what keeps a duplicate-only batch
-                # affordable. With no fresh landings and no adopted paths,
-                # ``restrict_files`` is None and this call walks the whole
-                # ``dest_folder`` — and re-importing a card into the same
-                # date folder it originally landed in means that folder
-                # already holds every twin. Non-incrementally that re-read
-                # and re-hashed all of them, turning a zero-copy import
-                # into an hours-long rescan on a network archive. Fresh
-                # landings are unaffected: they have no catalog row yet, so
-                # the skip path can't fire for them. Mirrors the dup-link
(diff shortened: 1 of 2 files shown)


jss367:
Good catch on the P2 — taking it. Pushed b0b0d8eb.

You're right that my "landed paths are always uncataloged" assumption doesn't hold. A stale row can survive for a file deleted off the archive, and if the replacement bytes land at that path with an mtime equal to the stale row's, the incremental fast path skips it without comparing size or content. `file_hash` keeps the stale value, the post-scan cross-check compares the copy-time hash against it, and a file that transferred fine gets reported failed — with retries unable to refresh the row. That's a stuck state, which is worse than the slow scan I was fixing.

Gated on `not scan_files`, as suggested:

```python
restrict_files=(scan_files or None),
incremental=not scan_files,
```

Worth noting this is strictly safer with no loss of the fix:

- **Sufficient** — the pathology is entirely in the `scan_files`-empty case, where `restrict_files` is None and the call walks the whole destination folder. `test_remote_import_dup_only_batch_does_not_reread_dest_folder_twins` still passes, so the hours-long rescan stays fixed.
- **Can't regress landings** — whenever `scan_files` is non-empty, `incremental` is now `False`, byte-identical to pre-PR behavior. So the failure mode you describe isn't just fixed, it's unreachable by this diff.
- **No upside lost** — with `restrict_files` narrowing those batches to a handful of paths, incremental was saving nothing there.

Tests: `test_import_job.py` 105 passed / 1 skipped; `test_scanner.py` + `test_import_dedup.py` + `test_jobs_api.py` 431 passed / 13 skipped.

PR #1404, opened 2026-07-31
Stop duplicate imports from scanning archive folders

jss367:
## Summary

- Link cataloged duplicate folders directly into the active workspace instead of scanning their archive directories.
- Skip remote duplicate-only batch scans while retaining exact-file scans for newly transferred or adopted files.
- Preserve `partial` folder status and leave stray-file repair to the explicit rescan workflow, while keeping workspace-link failures safety-gated.

## Why

Incremental duplicate-folder scans still enumerate and stat every entry, which can block zero-copy imports for hours on SMB-mounted NAS storage.

## Validation

`python -m pytest -q vireo/tests/test_import_job.py vireo/tests/test_jobs_api.py -k 'import' --maxfail=10` — 220 passed, 2 skipped.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of duplicate folders by linking existing catalog entries directly.
  * Recovered zero-byte files correctly instead of treating them as duplicates.
  * Promoted missing folders and reported linking failures more reliably.
  * Improved support for remote files, aliases, case-insensitive paths, mount changes, and stale cache data.
  * Prevented background sync jobs from interfering with subsequent operations.

### DIFF
--- a/vireo/import_job.py
+++ b/vireo/import_job.py
@@ -2057,13 +2057,6 @@ def _run_remote_import_job(job, runner, db, workspace_id, params):
                     if params.verify_by_hash:
                         verified -= 1
                     _fail(rel, dest_path, f"catalog scan failed: {e}")
-                # An adopted-only batch has no landed entries to roll back,
-                # but its pre-existing mount file still needs a catalog row.
-                if not landed and adopted_paths:
-                    _fail(
-                        rel, dest_folder,
-                        f"catalog scan failed for adopted mount files: {e}",
-                    )
                 landed = []
             else:
                 _invalidate_new_images(db, dest_folder)
--- a/vireo/import_job.py
+++ b/vireo/import_job.py
@@ -3106,12 +3106,13 @@ def run_import_job(job, runner, db_path, workspace_id, params):
                     if src_size == 0 and dest_size == 0:
                         # Zero-byte twin: identical by definition, but kept
                         # out of the duplicate-identity index (see ingest).
-                        skipped_duplicate += 1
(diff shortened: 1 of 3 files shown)


jss367:
🤖 CI red on `test-os (macos-latest)` — one test failed, unrelated to this PR's diff.

**Failure:**
```
vireo/tests/test_job_submission_api.py::test_job_sync_requests_serialize_xmp_work
Failed: job sync-1785529890324-2 did not reach phase 'Waiting for current XMP sync' within 30.0s; last_phase='Writing XMP metadata'
```

**Why I'm not pushing a fix:**

- The failing test is in `vireo/tests/test_job_submission_api.py`, which is **not** touched by this PR's own commits (`fd7f495`, `3c567ac`). This PR only changes `vireo/import_job.py` and `vireo/tests/test_import_job.py`. `test_job_submission_api.py`'s presence in the diff comes from the merge from main.
- The same test passed on `test-os (ubuntu-latest)` and `test-os (windows-latest)` on this same commit.
- Ran it 10× locally under `pytest -n 2` on Linux — all passing (~3s each).
- It's a threaded XMP-sync serialization test — timing-sensitive by nature. The observed `last_phase='Writing XMP metadata'` shouldn't happen while the first sync still holds `app._sync_job_lock` (blocked on `release_first`), so this looks like a macOS-specific runner timing quirk rather than a real regression.

**Recommended next step:** re-run just `test-os (macos-latest)`. If it flakes again, the fix belongs in a separate PR targeting the test's synchronization (e.g. blocking until the "Waiting" phase is observed via SSE rather than polling `runner.get()`).

---

PR #1442, opened 2026-08-09
Add portable computation cache

jss367:
## Summary

Adds an experimental portable computation cache so detector and species-classifier results can be exported as validated `.vireo-cache` bundles and reused on another Vireo installation by exact photo hash.
Introduces immutable local artifacts, runtime/input/full-label identities, secure bundle validation, catalog-later reuse, and review-aware replacement semantics that preserve decisions from every workspace.
Adds the Settings import/export UI, schema migration, design document, and coverage for bundle security, duplicate fan-out, idempotency, pinning, and HTTP round trips.

## Validation

- `vireo/tests/test_db.py`: 704 passed, 1 skipped
- `vireo/tests/test_pipeline_job.py`: 239 passed, 19 skipped
- `vireo/tests/test_classify_job.py`: 98 passed
- cache, schema, labels, Settings, and static UI suites: 124 passed

## Summary by CodeRabbit

- **New Features**
  - Added Portable Computation Cache controls in Settings.
  - Export detection and classification results as `.vireo-cache` bundles.
  - Import validated bundles and view processing summaries.
  - Reuse compatible cached results across installations and duplicate photos.

- **Improvements**
  - Cache reuse now accounts for model, runtime, input, and label versions.
  - Reuse cached results without loading models when possible.
  - Preserve reviews while reconciling taxonomy and grouping metadata.

- **Security & Reliability**
  - Validate bundles for corruption, unsafe paths, size limits, and integrity.
  - Quarantine results from unrecognized runtimes and apply imports transactionally.

### DIFF
--- a/docs/superpowers/specs/2026-08-08-portable-computation-cache-design.md
+++ b/docs/superpowers/specs/2026-08-08-portable-computation-cache-design.md
@@ -142,11 +142,12 @@ These are recommendations for review, not yet approved decisions.
 
 Settings gains a **Computation Cache** section with **Export Results…**.
 
-The export sheet offers:
+The phase-one controls offer:
 
-- scope: selected folders or all processed photos;
+- scope: all processed photos (selected-folder scope is follow-up work);
 - artifact types: Detection and Species classification (both selected by
-  default);
+  default); selecting Species classification automatically includes its
+  Detection dependency;
 - destination: a file ending in `.vireo-cache`.
 
 Before writing, Vireo reports how many photos have reusable results for the
@@ -156,8 +157,9 @@ counts would be misleading here. Export can hash missing non-empty photos after
 explicit confirmation; it does not silently read terabytes of originals and it
 does not repeatedly chase deliberately NULL hashes for empty files.
 
-The completion summary reports photos, detector runs, classifier runs, bundle
-size, skipped legacy results, and failures.
+The completion summary labels each unit explicitly: unique artifacts, affected
+catalog photo rows, detector runs, classifier runs, bundle bytes, artifacts
+stored for later, skipped legacy rows, and failures.
 
 ### Import on computer B
 
@@ -180,7 +182,8 @@ later, the next Process/Classify run discovers and applies the cached result.
 
 During processing, progress and the final job card show a line such as:
 
-> Reused 1,248 detector results and 1,103 classifications from imported cache.
+> Reused 1,248 detector-run artifacts across 1,310 catalog photo rows and
+> 1,103 classifier-run artifacts; stored 87 artifacts for later.
 
 The review UI does not need a permanent badge on every card. Prediction detail
 may show optional provenance: origin device label, Vireo version, execution
--- a/docs/superpowers/specs/2026-08-08-portable-computation-cache-design.md
+++ b/docs/superpowers/specs/2026-08-08-portable-computation-cache-design.md
@@ -1,0 +1,804 @@
+# Portable computation cache
+
+**Date:** 2026-08-08
+**Status:** Experimental phase-one implementation — format remains draft
+pending the cross-provider experiment
+
+## Implementation status
+
+The experimental phase-one bundle workflow is implemented behind an explicit
+**Experimental** label in Settings. It includes:
+
+- schema migration and full runtime/input/label identities for new detector
+  and classifier runs;
+- review-aware runtime replacement, including a pin from any workspace and an
+  explicit Reclassify override;
+- an immutable local object store and atomic `.vireo-cache` ZIP export;
+- full-bundle validation before import publication, including size, digest,
+  traversal, symlink, numeric-bound, and compression-ratio checks;
+- immediate exact-hash fan-out to matching catalog photos, plus local-store
+  reuse when a photo is cataloged after the bundle was imported;
+- Settings export/import controls and status summaries; and
+- automatic publication of newly computed portable detector and classifier
+  output when Vireo can prove the source, model, label, and input identities.
+
+Existing database results remain tagged `legacy`: they are valid local cache
+hits but are not exported. An explicit Reclassify recomputes them with the new
+identity contract. RAW+JPEG companion-backed runs also remain local-only until
+the canonical multi-source rendition contract is frozen.
+
+The cross-provider experiment, semantic-divergence diagnostics/tolerances,
+folder-scoped export sheet, retention controls, and shared NAS pack/index are
+still follow-up work. The artifact schema is therefore intentionally marked
+experimental rather than a compatibility promise.
+
+## Summary
+
+Let one Vireo installation reuse expensive detector and classifier work done by
+another installation, without sharing a live SQLite catalog and without
+turning Vireo into a cloud service.
+
+The first release is a portable cache bundle: computer A exports raw processing
+results for selected photos, computer B imports the bundle, and Vireo matches
+the results to its own catalog by the photo's SHA-256 content hash. A later
+release can point both computers at an append-only cache folder on a NAS and
+use the same artifact format automatically.
+
+This complements XMP rather than replacing it:
+
+- XMP remains the portable source of truth for accepted species, keywords,
+  ratings, flags, and other user decisions.
+- The computation cache carries reproducible machine output: detections,
+  classifier candidates, confidences, model identities, and completed-run
+  markers.
+- Workspace review state, edits, collections, and pending XMP changes remain
+  local unless another feature explicitly synchronizes them.
+
+This first release skips detector and classifier inference only. A full
+pipeline on computer B still computes working copies, masks, DINO embeddings,
+quality metrics, and any other enabled stages that are not yet portable.
+
+## Problem
+
+Wildlife photographers commonly keep originals on a NAS, process on a powerful
+desktop, and browse or review on a laptop. Vireo's current cache avoids repeated
+work across workspaces in one catalog, but it is local to `vireo.db`. A second
+computer that catalogs the same photo bytes cannot reuse the first computer's
+MegaDetector or classifier results.
+
+The current workarounds each lose something important:
+
+- Syncing accepted species to XMP shares the final decision, but not pending
+  predictions, alternatives, confidences, or detection boxes.
+- Copying `vireo.db` is a whole-catalog replacement, depends on compatible
+  absolute paths, and cannot safely merge two computers' work.
+- Putting the active SQLite database on a network share creates locking,
+  availability, and corruption risks and still does not define merge semantics.
+
+The useful first feature is therefore **shared computation**, not shared
+catalog state and not distributed job scheduling.
+
+## Goals
+
+1. Reuse detector and species-classifier results for byte-identical photos on
+   another Vireo installation.
+2. Keep the feature local-first: bundles and user-chosen filesystem locations,
+   with no Vireo-hosted service required.
+3. Never overwrite or implicitly transfer human review decisions.
+4. Reject stale results when the model, weights, canonical input rendition,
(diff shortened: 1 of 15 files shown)


jss367:
🤖 Reviewed the Codex round-8 findings against the reviewed commit (`17d83a3`) and each one appears to already be addressed by fixes that landed in earlier rounds. Rather than push a no-op commit, here's the evidence per finding:

| # | Codex finding (path:line) | Evidence in current tree |
|---|---|---|
| 1 | `pipeline_job.py:4235` — reapply classifications after detection | Post-detection `materialize_local_store` runs unconditionally at `pipeline_job.py:4519–4524` (inside the `if not params.reclassify` block starting at 4381), after synthetic full-image anchors are created at 4390–4478. |
| 2 | `classify_job.py:2511` — skip model resolution when cache satisfies | Peek at model is wrapped in `try/except` at `classify_job.py:2567–2603`; `_all_photos_cache_satisfied` gate at 2636 returns via `_finalize_cached_only` at 2649 before `get_active_model()` at 2668. |
| 3 | `computation_cache.py:1481` — reject unknown detector runtimes | `_is_recognized_detector_runtime` is called at `computation_cache.py:1447` before the write; body at 1286–1316 requires the fingerprint to match a locally-derivable runtime (or the persisted trust set). |
| 4 | `computation_cache.py:585` — validate candidate taxonomy | `_validate_candidate_taxonomy` is called at 586; body at 471–490 rejects non-dict taxonomy and non-string scalar fields (`genus`, `kingdom`, …). |
| 5 | `classify_job.py:2640` — match requested classifier before short-circuit | `_all_photos_cache_satisfied` at 379–384 filters on `classifier_model` and `labels_fingerprint`; 457–459 requires `covered == total` across all detections; identity_gate at 401/441–455 rejects mixed identities. |
| 6 | `computation_cache.py:1664` — verify classifier runtimes | `_is_recognized_classifier_runtime` is called at `computation_cache.py:1516` before the classifier write; body at 1362–1400 reproduces the fingerprint from a locally-downloaded classifier (or the trust set). |
| 7 | `computation_cache.py:601` — validate label metadata scalars | `labels.display_name` bounded string check at 616–622; `labels.count` non-negative int64 check at 623–639. |
| 8 | `app.py:16701` — persist trust for unmatched imports | `store.record_trusted_runtimes(...)` at `app.py:16694–16697` persists to `trust.json`; `materialize_local_store` unions the persisted set at `computation_cache.py:1691–1697` on every subsequent call. |
| 9 | `classify_job.py:2573` — reject unknown explicit model IDs | `model_id_missing = True` set at `classify_job.py:2584` when the requested `params.model_id` is not in `get_models()`; raises `RuntimeError("Model '…' not found or not downloaded.")` at 2610–2613 before `_all_photos_cache_satisfied` runs at 2636. |

If any of these are actually incorrect (e.g. I'm quoting a code path that isn't reached in the scenario Codex has in mind), please point at the specific input/scenario and I'll dig further and push a fix.

---

jss367:
Codex's latest review on commit `06a9f31f83` contained only the boilerplate header with no inline suggestions attached, so there is nothing actionable to address. All prior Codex inline comments were on earlier commits and have already been handled in this PR's history. No changes pushed.

---

jss367:
🤖 Codex round-13 review (commit `8baa83bc3e`) — status:

**Finding 1 — `computation_cache.py:219` "Include taxonomy identity in classifier runtimes"** (P1): addressed in commit `2e6bd255`. `TimmClassifier` consults `taxonomy.json` for common-name resolution (fallback when `label_descriptions.json` misses a class) and for the `taxonomy` hierarchy fields stored on every prediction — both land in the published artifact, so two installations with different taxonomy revisions produce semantically distinct output for the same raw class id. The `classifier_runtime_fingerprint` formula now includes a portable content identity (SHA‑256 of the backing taxonomy file, or `no-tax` when none is loaded), threaded through the publish path, the standalone Classify peek‑cache‑check, the post‑detect known‑runtime whitelist, and the built‑in `_is_recognized_classifier_runtime` fallback. Bundles produced under a different taxonomy now stay quarantined at materialize time until the destination reproduces the same identity, matching the model/labels/detector‑runtime quarantine policy the design doc already spells out.

**Finding 2 — `pipeline_job.py:4235` "Apply pipeline cache before requiring model weights"** (P1): the finding is valid but not addressed in this push, and I'd like to align on scope before pushing a partial fix. Analysis and proposed direction:

- **Root cause is architectural, not local to line 4235.** The pipeline runs `scanner`, `collection`, `thumbnail`, and `model_loader` concurrently in phase 1 (`pipeline_job.py:7758–7786`), then `detect` (which contains the current `materialize_local_store` call at 4231–4241 and the post‑detect reapply at 4519–4540). `model_loader_stage` fails fatally if the requested model isn't downloaded — `_load_model_bundle` raises `RuntimeError("<model>: install the missing files via Repair.")` when `_classify_model_state` isn't `ok`/`unverified` (`pipeline_job.py:3800–3808`), and `resolution_error` from the pre‑phase resolve (1622–1637) is re‑raised at 4051–4052. Either path calls `abort.set()`, which makes `detect_stage` no‑op (skipped) before its `materialize_local_store` ever runs. The standalone flow already handles this cleanly (peek → cache check → `_finalize_cached_only`), but the pipeline can't just "call the cache check earlier" — it also has to have somewhere to reconcile the reused rows, since `classify_stage`'s per‑model loop is built around iterating detections through `_flush_batch` / `_prepare_image`.

- **What a proper fix looks like** (larger than a single review round IMO):
  1. Split `model_loader_stage` into "resolve spec (may be undownloaded)" and "load bundle (needs weights)".
  2. Move the initial `materialize_local_store` to phase 1 (or an early sub‑step of `detect_stage`) so cache is applied before `_load_model_bundle` decides whether to raise.
  3. Add `loaded_models["cache_only_mode"]` set when the pipeline’s collection is fully cache‑satisfied for every requested `(model, labels_fingerprint, taxonomy_identity)` triple, and demote the "not downloaded" error to a soft skip in that case.
  4. Give `classify_stage` a cache‑only branch that mirrors `_finalize_cached_only`: reconcile cached predictions via `_store_grouped_predictions` (so imported rows still get auto‑match / burst grouping) without invoking `clf`.
  5. Multi‑model runs need per‑spec cache‑coverage decisions — a mixed collection where model A is fully cached but model B isn't should still process model B normally instead of aborting on the missing weights.

- **Why I'm asking before pushing that:** step 4 touches the classify per‑model loop, contextual weak rescue, and the multi‑spec bundle‑release/`_release_classifier_cache_handle` bookkeeping. A minimal patch that flips only step 3 without also doing 4 would trade "hard fail" for "silent success without reconciliation" (imports stay categorized `new`, no group/consensus metadata) — which is exactly the trap Codex flagged and the maintainer fixed in the standalone flow via `_finalize_cached_only` in earlier rounds. Would you like me to proceed with the full four‑step change as its own commit on this PR, or split it into a separate PR so this one can settle on the taxonomy + earlier‑round fixes first?

---

jss367:
🤖 Reviewed the Codex round-19 findings against the reviewed commit (`deb7d99b`) — every unresolved thread already has a guard in the current tree; no code changes pushed. Evidence per finding (P1 first):

| # | Codex finding (path:line) | Evidence in current tree |
|---|---|---|
| 1 | `pipeline_job.py:4269` — reapply classification artifacts after detection | Post-detection reapply lives in the `if not params.reclassify:` block at `pipeline_job.py:4415-4585`. Empty-scene full-image anchors are pre-created at 4432-4512, `known_runtimes` + `known_classifier_runtimes` are assembled at 4514-4557, and `materialize_local_store(...)` is called at 4563-4569 — **before** `classify_stage` runs at 4610. The standalone Classify flow has the mirroring post-detection call at `classify_job.py:3324` (block 3192-3341). The pre-detection call at 4269 is intentionally the first pass; the second pass at 4563 is what catches classification-only bundles. |
| 2 | `classify_job.py:2658` — skip model resolution when cache satisfies | Model peek at `classify_job.py:2705-2765` uses catalog lookup + `get_active_model()` **without loading weights**, wrapped in try/except so a fresh install with no downloaded weights yields `peek_model=None` instead of raising. `_all_photos_cache_satisfied` runs at 2855; on a hit `_finalize_cached_only` returns at 2872 **before** the hard "No model available" gate at 2891-2892. |
| 3 | `computation_cache.py:1778` — reject unknown detector runtimes | `_is_recognized_detector_runtime(...)` is called at `computation_cache.py:1744-1747` and returns False for unknown fingerprints, incrementing `result["unknown_runtime"]` and `continue`ing — the write at 1772 is unreachable for foreign runtimes. Body at 1468 reproduces the runtime locally (or accepts a caller-supplied `known_runtimes` whitelist). |
| 4 | `classify_job.py:2863` — match requested classifier before short-circuit | `_all_photos_cache_satisfied` (335-537) receives `desired_classifier_model`, `desired_labels_fingerprint`, `desired_labels_fingerprint_full`, `desired_model_identity`, and `cache_taxonomy_identity` (resolved at 2705-2854) and SQL-filters by them at 395-450. Coverage requires **every** classifiable detection covered (`if covered != total: return False` at 535), not "one per photo". A mixed-identity collection is also rejected at 510-533. |
| 5 | `computation_cache.py:1961` — verify classifier runtimes before materializing predictions | `_is_recognized_classifier_runtime(...)` is called at `computation_cache.py:1813-1820` **before** the labels_fingerprints insert (1846+) and any per-subject write. Failure increments `result["unknown_classifier_runtime"]` and `continue`s. The body at 1544-1595 reproduces the classifier runtime from installed weights + taxonomy identity + labels fingerprint, or accepts the caller's `known_classifier_runtimes` whitelist. Detector-dependency check at 1826-1841 runs additionally, not instead. |
| 6 | `app.py:16701` (via ordinal drift) — persist trust for unmatched imports | `record_trusted_runtimes(...)` persists to `trust.json`; `materialize_local_store` unions the persisted set on every subsequent call (previously walked through at rounds 8, comment 5230811473). No regression here. |
| 7 | `classify_job.py:2863` — finalize imported results before returning | The cache short-circuit at 2872-2876 calls `_finalize_cached_only` (2307-2475), which fetches detections (2349-2375), reads cached predictions (2393-2398), and calls `_store_grouped_predictions(...)` at 2446 — the same finalize path used by fresh classification. Auto-match reconciliation, group_id, burst grouping, and consensus metadata all land. |
| 8 | `classify_job.py:2720` — reject unknown explicit model IDs before cache reuse | When `params.model_id` isn't in `get_models()`, `model_id_missing = True` is set at `classify_job.py:2731` (or 2759 for the transient-failure branch) and raises `RuntimeError("Model '…' not found or not downloaded.")` at 2772-2775 — **before** `_all_photos_cache_satisfied` runs at 2855. |
| 9 | `computation_cache.py:677` — validate candidate taxonomy | `_validate_candidate_taxonomy` at `computation_cache.py:563-582` rejects a non-dict taxonomy (`taxonomy = ["…"]`) at 571-572 and rejects any listed scalar field (`taxonomy.genus = ["…"]`) via `isinstance(value, str)` at 579. Called during the initial all-or-nothing pass at 678, so `import_bundle` fails before any object is published (the round-19 commit `deb7d99b` additionally stages every object body to a temp dir so a later per-object or manifest failure leaves the store untouched). |
| 10 | `computation_cache.py:693` — validate scalar label metadata | `labels.display_name` string+length check at 717-723; `labels.count` int + non-negative + SQLite INT64 range check at 724-740 (explicitly rejects `bool` and OverflowError-risk values). Both run inside `validate_artifact`, i.e., during the initial bundle pass. |

If any of these misread the scenario Codex has in mind, please point at the specific input / call path and I'll dig further and push a fix.

---

PR #1446, opened 2026-08-09
Prevent SMB photo deletes from hanging

jss367:
Vireo could hang indefinitely while moving photos to Trash on an unhealthy macOS SMB mount because `os.replace` ran inside the server worker.
This routes network-volume deletions through the existing time-bounded Finder subprocess, fails closed when mount discovery fails, retains catalog rows on timeout or error, and preserves the atomic rename fast path for local removable volumes.
It also reports file-level batch-delete progress and adds regression coverage for mount classification, bounded routing, failure behavior, and the progress callback.
Validation: 42 delete API tests, 13 focused Trash tests, and 32 duplicate-cleanup tests pass; the full app module has 426 passes and one unrelated local ExifTool availability failure.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved trash handling for files on macOS network volumes by routing them through Finder.
  * More reliably distinguishes moved, missing, and failed items, including files that are already missing.
  * Prevents unsafe deletion when network-volume status cannot be confirmed.
  * Handles custom and symlinked network mount points more reliably.
  * Improves duplicate-file cleanup for network-stored files and removes corresponding orphaned records.

* **Improvements**
  * Added per-item progress reporting for trash and permanent deletion operations.
  * Provides clearer outcomes when files disappear or network mounts change during processing.

### DIFF
--- a/vireo/app.py
+++ b/vireo/app.py
@@ -1725,20 +1725,117 @@ def _network_volume_roots(run=subprocess.run):
     }
 
 
+def _expand_first_symlink_prefix(filepath):
+    """Expand one local symlink prefix without resolving its target.
+
+    ``realpath`` follows the target and can block when that target is an
+    unhealthy network share. Reading the first symlink itself only touches
+    its local directory entry; the unvisited suffix is then appended
+    lexically so network-volume classification stays free of share I/O.
+    """
+    try:
+        normalized = os.path.normpath(os.path.abspath(filepath))
+    except (OSError, TypeError, ValueError):
+        return None
+    parts = normalized.split(os.sep)
+    prefix = os.sep
+    for index, part in enumerate(parts[1:], start=1):
+        if not part:
+            continue
+        prefix = os.path.join(prefix, part)
+        try:
+            target = os.readlink(prefix)
+        except OSError:
+            continue
+        if not os.path.isabs(target):
+            target = os.path.join(os.path.dirname(prefix), target)
+        return os.path.normpath(os.path.join(target, *parts[index + 1:]))
+    return None
+
+
 def _path_on_network_volume(filepath, network_roots):
     """Whether ``filepath`` should avoid in-process mounted-volume I/O."""
-    if network_roots is None:
-        # Mount discovery failed.  Prefer the bounded Finder path for any
-        # mounted volume to an os.replace() that can wedge the server forever.
-        return _volume_root_for_path(filepath) is not None
     normalized = os.path.normpath(os.path.abspath(filepath))
-    for root in network_roots:
-        try:
-            if os.path.commonpath((normalized, root)) == root:
+    for _depth in range(16):
+        if network_roots is None:
+            # Mount discovery failed. Prefer bounded Finder handling for any
+            # /Volumes path, including one reached through a local symlink.
+            if _volume_root_for_path(normalized) is not None:
                 return True
-        except ValueError:
-            continue
-    return False
+        else:
+            for root in network_roots:
+                try:
+                    if os.path.commonpath((normalized, root)) == root:
+                        return True
+                except ValueError:
+                    continue
+        if sys.platform != "darwin":
+            return False
+        expanded = _expand_first_symlink_prefix(normalized)
+        if expanded is None:
+            return False
+        normalized = expanded
+    # A symlink loop or unusually deep chain cannot be classified safely.
+    # Fail closed on macOS so no subsequent stat reaches a possible share.
+    return True
+
+
+def _missing_paths_via_finder(filepaths, timeout=_FINDER_TRASH_TIMEOUT_SECS):
+    """Boundedly confirm which paths remain absent according to Finder."""
+    filepaths = [os.fspath(path) for path in filepaths]
+    if not filepaths:
+        return set(), set(), []
+    result = subprocess.run(
+        [
+            "osascript",
+            "-e", "on run argv",
+            "-e", "set statuses to {}",
+            "-e", "repeat with posixPath in argv",
+            "-e", "set statusValue to \"error\"",
+            "-e", "try",
+            "-e", "set fileRef to POSIX file (contents of posixPath)",
+            "-e", "tell application \"Finder\"",
+            "-e", "if exists fileRef then",
+            "-e", "set statusValue to \"exists\"",
+            "-e", "else",
+            "-e", "set statusValue to \"missing\"",
+            "-e", "end if",
+            "-e", "end tell",
+            "-e", "end try",
+            "-e", "set end of statuses to statusValue",
+            "-e", "end repeat",
+            "-e", "set AppleScript's text item delimiters to linefeed",
+            "-e", "return statuses as text",
+            "-e", "end run",
+            "--",
+            *filepaths,
+        ],
+        capture_output=True,
+        text=True,
+        timeout=timeout,
+        **no_window_kwargs(),
+    )
+    if result.returncode != 0:
+        raise OSError(
+            result.stderr.strip()
+            or f"Finder existence check failed ({result.returncode})"
+        )
+    statuses = result.stdout.splitlines()
+    if len(statuses) != len(filepaths):
+        raise OSError("Finder returned an invalid existence response")
+    missing = set()
+    existing = set()
+    failures = []
+    for filepath, outcome in zip(filepaths, statuses, strict=True):
+        if outcome == "missing":
+            missing.add(filepath)
+        elif outcome == "exists":
+            existing.add(filepath)
+        else:
+            failures.append({
+                "path": filepath, "error": "Finder existence check failed",
+            })
+    return missing, existing, failures
 
 
 def _ensure_volume_trashes_dir(filepath, ensured_volumes):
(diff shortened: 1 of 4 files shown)


jss367:
Both Codex findings are addressed by 25b772f6:

- **P1** (network classification ahead of preflight I/O) — `_trash_paths` now partitions network paths into `finder_candidates` before `_snapshot_parent_device` / `os.path.isfile` run, so a stalled SMB mount can no longer pin a worker on the in-process stat that used to precede the network-volume branch.
- **P2** (mounts outside `/Volumes`) — `_path_on_network_volume` no longer early-returns `False` when `_volume_root_for_path` is `None`; it now compares against the discovered network roots directly, so SMB/NFS mounts at custom mount points (e.g. `/Users/me/mnt/photos`) still route through the bounded Finder subprocess. Fail-closed behavior is preserved for `/Volumes` paths when the mount table can't be read.

Related trash/network tests pass locally (10/10 in `vireo/tests/test_app.py`). No additional changes from the agent.

---

jss367:
Both P2 Codex findings on `b292a15b` are addressed by `7ed82b68`:

- **P2** (already-missing network losers) — `_trash_paths` gained an opt-in `already_missing_out` set; the delete-loser-files endpoint passes it in and surfaces those paths as `{reason: "file already missing"}` in `skipped`, so a network loser that is already gone on a still-mounted share now returns the same terminal outcome the local branch has always returned. `trashOneLoserFile`'s existing "file already missing" branch fires and the card clears instead of sitting on "Moving to Trash..." forever.
- **P2** (per-path mount subprocesses) — `_finder_missing_is_trustworthy` no longer calls `_network_volume_roots()` itself; the batch loop queries the mount table at most once per Finder batch (and only when the batch actually contains a network `missing` outcome) and passes the shared roots into the trust check for every path. A 20-item all-missing retry now takes one `mount` subprocess, not twenty.

Regression coverage:
- `test_trash_paths_reuses_one_mount_recheck_per_finder_batch` asserts the mount table is queried only twice for a 20-path all-missing Finder batch (initial classification + single per-batch recheck).
- `test_trash_paths_exposes_already_missing_paths_via_out_parameter` asserts the new set is populated only for paths whose end state already held, not for paths actually moved.
- `test_delete_loser_files_reports_already_missing_network_losers_as_skipped` asserts the endpoint surfaces the network-already-missing case as `{"reason": "file already missing"}` and still drops the orphan DB row.

Related trash/duplicates suites pass locally (19/19 trash tests in `vireo/tests/test_app.py`, all `vireo/tests/test_duplicates_api.py` and `vireo/tests/test_delete_api.py`). The pre-existing unrelated exiftool-availability failure in `test_failed_import_does_not_chain` is not caused by this change (verified against `b292a15b`).

---

jss367:
The P1 Codex finding on `e4fcf678` ("Probe the deepest matching network mount") is addressed by `07241e44`:

- Introduced `_deepest_network_root_for_path`: longest-common-prefix matching over `network_roots`, with the same one-symlink-prefix expansion loop as `_path_on_network_volume`, returning the deepest matching root.
- `_trash_paths` now builds a per-batch `path_to_deepest_root` mapping, probes reachability of each distinct deepest root, and passes the mapping into `_finder_missing_is_trustworthy`. Trust now requires the path's *deepest* matching root to be in `reachable_network_roots` — a reachable outer share can no longer vouch for a nested detached inner share (e.g. `/Volumes/NAS/archive` beneath `/Volumes/NAS`). The old code iterated `batch_network_roots` (a set) with a first-match `break`, so which root got probed was set-iteration-order-dependent.

Regression coverage:
- `test_trash_paths_rejects_missing_when_nested_inner_mount_unreachable` sets up two paths on nested mounts (outer reachable, inner detached); asserts the outer path is accepted as missing, the inner is preserved for retry, and `_network_root_reachable` was called for *both* deepest roots — proving we no longer rely on set iteration order picking a single winner.

The other Codex findings on earlier commits (line 1798 symlink, line 2042 mount-identity, line 2189 duplicate-cleanup preflight, line 22487 already-missing losers, line 22519 preserve endpoint classification) were already addressed by later commits in this branch — verified against current HEAD.

Full trash/network suite passes (28/28 in `vireo/tests/test_app.py`). Broader mandated suite: 2097 pass, 16 skip; the pre-existing unrelated `test_failed_import_does_not_chain` still fails only because `exiftool` isn't available in this environment (same failure noted on `b292a15b`).

---

jss367:
The P2 Codex finding on `4966747` ("Bound reachability probes across the whole Finder batch") is addressed by `508773e`:

- **P2** — `_trash_paths` now probes each distinct deepest network root via `concurrent.futures.ThreadPoolExecutor` instead of a serial `for` loop. Each `_network_root_reachable()` still runs its own bounded `stat` subprocess with `_MOUNT_QUERY_TIMEOUT_SECS`, but wall-clock is now bounded by one probe timeout rather than `len(distinct_roots) × timeout`. A full 20-item Finder batch spanning unreachable shares completes its reachability sweep in ~5s instead of up to ~100s, so the batch's Finder recheck can no longer be starved by the pre-recheck reachability loop.

Regression coverage:
- `test_trash_paths_probes_distinct_roots_concurrently` uses a `threading.Barrier` sized to the number of distinct roots (6). A serial implementation would never satisfy the barrier's party count and `wait()` would raise `BrokenBarrierError` on the 5-second timeout — the test asserts the trash run completes successfully, which only holds when the probes execute concurrently.
- The pre-existing `test_trash_paths_reuses_one_mount_recheck_per_finder_batch` (single root) and `test_trash_paths_rejects_missing_when_nested_inner_mount_unreachable` (multi-root correctness) continue to pass, so the concurrency change did not regress single-root or nested-root behaviour.

The remaining unresolved Codex threads on earlier commits (line 1807 symlink, line 2096 mount-identity, line 2243 duplicate-cleanup preflight, line 22556 already-missing losers, line 22590 preserve endpoint classification) were already addressed by later commits — verified against current HEAD.

Full mandated suite: 2099 pass, 16 skip. The pre-existing unrelated `test_failed_import_does_not_chain` still fails only because `exiftool` is not available in this environment (same environmental failure noted on `b292a15b` / `e4fcf678`, reproduces on the pre-change commit `4966747`).

---

jss367:
The three P1 Codex findings on `508773ed` are addressed by `b3a8197`:

- **P1** (avoid synchronously reaping a wedged stat process) — `_network_root_reachable` no longer calls `subprocess.run(..., timeout=...)`; it drives `Popen` directly and, on timeout, hands the child to the new `_abandon_stuck_subprocess` helper, which `SIGKILL`s the process group and reaps on a daemon thread. `subprocess.run`'s post-kill `wait()` would keep the caller blocked as long as `/usr/bin/stat` was stuck in an uninterruptible SMB kernel call — the exact case this probe exists to detect — and because the reachability sweep runs inside a `ThreadPoolExecutor`, the executor context then blocked on the same worker.
- **P1** (preserve detached network mounts as network paths) — `_path_on_network_volume` now fails closed on any macOS `/Volumes/...` path regardless of the discovered set. Previously, a NAS that had detached before mount discovery ran left its stub mount-point directory in place while `mount` returned no matching root, so `/Volumes/NAS/photo.jpg` fell through to the local branch; `_snapshot_parent_device` and `os.path.isfile` then succeeded against the stub directory, and callers pruned catalog rows for photos that reappear on remount. A `/Volumes` path missing from the current mount table is more plausibly a detached share than a genuinely local `/Volumes` subdirectory, so we route it through Finder.
- **P1** (fail closed for custom mounts when discovery fails) — the `network_roots is None` fallback previously protected only `/Volumes/...`, leaving supported custom mount locations like `/Users/me/mnt/photos` free to hang the caller on an in-process stat. Now every macOS candidate is treated as network-backed when discovery itself failed — we cannot distinguish a custom network mount from an ordinary local path, and the fail-closed cost (some Finder overhead on genuinely local files during a rare discovery failure) is smaller than the unbounded-hang cost on an unhealthy share.

Regression coverage:
- `test_network_root_reachable_abandons_child_that_hangs_on_kill` simulates the uninterruptible-kernel-call case with a fake `Popen` whose `wait()` blocks forever, then asserts the probe returns within a 5s deadline from a worker thread — the previous synchronous `wait()` path would fail this test in production against any wedged SMB mount.
- `test_path_on_network_volume_fails_closed_on_detached_volumes_mount` asserts `/Volumes/NAS/photo.jpg` is classified as network-backed even when the discovered roots either omit it or are entirely empty.
- `test_path_on_network_volume_fails_closed_for_custom_mounts_when_discovery_fails` asserts `/Users/me/mnt/photos/bird.NEF` is classified as network-backed when `network_roots is None` on macOS, and remains local on non-macOS platforms.

The three existing `test_network_root_reachable_*` tests were updated to inject a fake `Popen` (via the new `_popen` kwarg) rather than a fake `subprocess.run`. `test_network_mount_discovery_failure_avoids_direct_volume_rename` gained an explicit `sys.platform="darwin"` monkeypatch since the `/Volumes` fail-closed fallback is now correctly darwin-gated — no bounded Finder alternative exists on Linux/Windows, so classifying a path as network-backed there would just make callers fail the operation for no benefit.

Full `test_app.py` suite passes (451/451), plus `test_duplicates_api.py` and `test_delete_api.py` (73 pass, 3 skip). Mandated suite excluding `test_app`: 1651 pass, 16 skip. The pre-existing unrelated `test_failed_import_does_not_chain` still fails only because `exiftool` is not available in this environment (same environmental failure noted on `4966747` / `508773e`).

---

PR #1448, opened 2026-08-10
Fail closed on unreachable macOS network mounts

jss367:
Follow-up to #1446 addressing three late P1 review findings after that PR was automatically merged. Mount discovery failures now route all macOS candidates through bounded Finder handling, and detached paths under /Volumes remain network-classified even when the current mount table no longer lists them. Network-root stat probes now use Popen and hand timed-out children to a daemon reaper instead of synchronously waiting on an uninterruptible filesystem call. Regression tests cover wedged-probe abandonment, detached mount-table entries, and custom mount discovery failure.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved network reachability checks so timed-out or failed probes return promptly and clean up reliably.
  * Limited and deduplicated network checks to improve responsiveness and stability.
  * macOS network-volume detection now safely handles undiscoverable, detached, or unverified mounts.
  * Trash operations avoid unintended local filesystem changes and incorrect success reports for network paths.

* **Tests**
  * Added coverage for stalled probes, asynchronous cleanup, concurrent probe limits, network-volume detection, and safe trash behavior.

### DIFF
--- a/vireo/app.py
+++ b/vireo/app.py
@@ -1735,8 +1735,30 @@ def _network_volume_roots(run=subprocess.run):
     }
 
 
+def _reap_abandoned_network_probe(process):
+    """Reap a timed-out probe away from the request path."""
+    try:
+        process.communicate()
+    except (OSError, ValueError, subprocess.SubprocessError):
+        pass
+
+
+def _abandon_network_probe(process):
+    """Kill a timed-out probe without synchronously waiting for it."""
+    try:
+        process.kill()
+    except OSError:
+        pass
+    threading.Thread(
+        target=_reap_abandoned_network_probe,
+        args=(process,),
+        name="vireo-network-probe-reaper",
+        daemon=True,
+    ).start()
+
+
 def _network_root_reachable(root, timeout=_MOUNT_QUERY_TIMEOUT_SECS,
-                             run=subprocess.run):
+                             run=None, popen=subprocess.Popen):
     """Bounded, out-of-process reachability probe for a mounted network root.
 
     ``mount`` listing a share and Finder reporting a file's absence are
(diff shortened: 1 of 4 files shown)


jss367:
The Codex P1 (`Preserve local volumes in the local-trash path`, `vireo/app.py:1867`) is already resolved on this branch by commits `abed1be3` (`Preserve discovery fallback on test hosts`) and `7d5ede5c` (`Preserve live local volume trash routing`), which introduce `_NetworkVolumeRoots(set)` carrying the live top-level `/Volumes/<name>` mount points from the same `mount` snapshot so `_path_on_network_volume` can distinguish healthy local USB/APFS/ExFAT disks from detached network mount stubs. Regression coverage in `test_path_on_network_volume_preserves_live_local_volume` and `test_trash_paths_preserves_live_local_volume_fast_path` passes locally. No further push needed.

---

jss367:
Addressed the outstanding Codex P1 on `vireo/remote_setup.py:40` (`Stop treating autofs mounts as local`) in `b5e26ebb`: `autofs` is now excluded from `_LOCAL_FS`, so `mount_type_is_network_or_unknown("autofs")` returns True and `_path_on_network_volume` routes autofs paths through the bounded, out-of-process reachability probe instead of an in-process `stat`/`isfile` that would trip the automount. `parse_mount_output` is unaffected (network-share listing still uses the positive `_NETWORK_FS` allowlist). New regression assertion in `test_unknown_mount_types_fail_closed_as_possibly_network_backed`.

The Codex P2 (`Cap abandoned reachability probes`, `vireo/app.py:1808`) is already addressed on the current HEAD by `3afeaa5c` (`Bound stuck network reachability probes`): `_reserve_network_probe` bounds concurrent probes with `_MAX_NETWORK_PROBES` and per-root deduplication in `_NETWORK_PROBES`, so repeated Trash retries against a wedged mount reuse the reservation instead of spawning a new probe + reaper per attempt.

Local run of the required suite (`tests/test_workspaces.py vireo/tests/test_db.py vireo/tests/test_app.py vireo/tests/test_photos_api.py vireo/tests/test_edits_api.py vireo/tests/test_jobs_api.py vireo/tests/test_darktable_api.py vireo/tests/test_config.py`) is green except `test_failed_import_does_not_chain`, which fails identically on the pre-change commit `c334ec84` in this container — the test relies on `os.chmod(path, 0)` to make a file unreadable, but the test runs as root here so the chmod is ignored. Unrelated to this fix.

---

PR #1453, opened 2026-08-11
Scope card cleanup verification to matching files

jss367:
The card cleanup flow now verifies only archive copies matching the selected card, rather than re-hashing the entire workspace. It atomically refreshes the existing preview, deduplicates checks by content hash, and leaves unrelated archive files untouched. Deletion now reads each card file once while retaining fresh archive checks around that read and all existing mutation and overlap guards. Validation: 94 focused unit, API, and browser tests pass, along with the missing-originals heavy-job regression, Ruff, JavaScript syntax, and diff checks.

## Summary by CodeRabbit

- **New Features**
  - Added targeted archive verification for card cleanup, checking only matching archive copies.
  - Added progress, cancellation, and completion reporting for verification jobs.
  - Cleanup previews now refresh automatically with updated manifests and file status details.

- **Bug Fixes**
  - Improved deletion safety with archive rechecks and protection against outdated previews.
  - Prevented overlapping or stale verification and deletion operations.
  - Reduced unnecessary card-file reads during deletion.
  - Improved handling of unavailable, changed, corrupt, or unreadable archive copies.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,9 @@ All notable changes to Vireo are documented in this file.
   updater, and uninstall-preservation gates before publication.
 
 ### Fixed
+- Freeing card space now verifies only archive copies that match the selected
+  card instead of re-hashing the entire workspace, refreshes the preview
+  automatically, and reads each card file only once at deletion.
 - Opening a catalog that was already migrated by a newer Vireo (for example
   after reinstalling an older build) now shows clear "update Vireo to open
   this catalog" guidance instead of a crash — and no longer suggests moving
--- a/docs/superpowers/specs/2026-08-07-card-cleanup-design.md
+++ b/docs/superpowers/specs/2026-08-07-card-cleanup-design.md
@@ -293,13 +293,20 @@ photo) are both deletable — the rule is content-based, not one-to-one.
 
 ### Delete job
 
-`POST /api/card-cleanup/delete` with `{scan_job_id}` starts
-`card_cleanup_delete`. It loads the manifest file written by the scan and
-refuses to start if the scan job is missing, unfinished, or cancelled, if
-the manifest file is gone (expired), or if its deletable bucket is empty.
-Only one delete job per scan manifest may run at a time. Because the
-manifest lives on disk, delete-after-restart works: the endpoint validates
-the scan job against `job_history` when it is no longer in memory.
+`POST /api/card-cleanup/delete` with `{scan_job_id, manifest_revision}`
+starts `card_cleanup_delete`. `manifest_revision` is the integer revision
+the client last rendered (from the manifest's `revision` field); a
+targeted verify job atomically bumps this after each run, so a mismatch
+means the on-disk preview no longer matches what the user confirmed.
+The endpoint responds `409 manifest_revision_mismatch` in that case so
+the UI can drop the user back into a "reload the preview" state rather
+than sweep in newly-promoted files the user never saw. It otherwise
+loads the manifest file written by the scan and refuses to start if the
+scan job is missing, unfinished, or cancelled, if the manifest file is
+gone (expired), or if its deletable bucket is empty. Only one delete
+job per scan manifest may run at a time. Because the manifest lives on
+disk, delete-after-restart works: the endpoint validates the scan job
+against `job_history` when it is no longer in memory.
 
 For each **deletable** manifest entry, immediately before deletion:
 
@@ -352,11 +359,18 @@ they exist only for the preview.
   matching, previously-unchecked archive copies and atomically refreshes the
   scan manifest. Returns the job id; 409 if verification or deletion for the
   same manifest is already running.
-- `POST /api/card-cleanup/delete` — body `{scan_job_id}`. Returns the job
-  id. 409 if a delete for that manifest is already running; 400 for an
-  unfinished/cancelled scan or a manifest that fails load-time validation
-  (corrupt, wrong schema, entries outside the source root); 404 for an
-  unknown scan job, a pruned manifest file, or a manifest older than 7
+- `POST /api/card-cleanup/delete` — body
+  `{scan_job_id, manifest_revision: int}`. Returns the job id. 400 if
+  `manifest_revision` is missing or not an integer (bools are rejected
+  explicitly — `isinstance(True, int)` is truthy). 409
+  `manifest_revision_mismatch` when the on-disk manifest carries a
+  different revision than the client confirmed (a verify run has since
+  refreshed the preview); the UI reloads and asks the user to
+  re-confirm. Otherwise 409 if a delete for that manifest is already
+  running; 400 for an unfinished/cancelled scan or a manifest that
+  fails load-time validation (corrupt, wrong schema, entries outside
+  the source root); 404 for an unknown scan job, a pruned manifest
+  file, or a manifest older than 7
   days (age is checked at request time, not only at prune time), with
   "re-scan the card" copy.
 - Progress and results ride the existing job endpoints (stream, status,
--- a/docs/superpowers/specs/2026-08-07-card-cleanup-design.md
+++ b/docs/superpowers/specs/2026-08-07-card-cleanup-design.md
@@ -173,9 +173,11 @@ verified files** button opens a confirmation dialog that states plainly:
 - Deletion is permanent — memory cards have no trash.
 - After deletion, the archive holds the only copy of these photos.
 - What "verified" means: each file's archive copy passed a checksum check
-  (at import, targeted cleanup verification, or integrity audit) and is
-  confirmed unchanged since by size and timestamp; the card copy itself is
-  re-checksummed at the moment of deletion. Archive bytes are not re-downloaded.
+  (at import, targeted cleanup verification, or integrity audit) and its
+  current size and modification time still match the `file_size` /
+  `file_mtime` baseline the catalog recorded when the checksum was
+  written; the card copy itself is re-checksummed at the moment of
+  deletion. Archive bytes are not re-downloaded.
 
 Confirming starts the **delete** job with live per-file progress. The final
 summary reports exact counts: deleted, kept, skipped-because-changed, and
--- a/docs/superpowers/specs/2026-08-07-card-cleanup-design.md
+++ b/docs/superpowers/specs/2026-08-07-card-cleanup-design.md
@@ -11,7 +11,7 @@ manifests); awaiting maintainer sign-off before implementation planning
 card) only after verifying, per file and at deletion time, that the card
 file's bytes match a hash that was checksum-verified into the archive and
 that the archive copy is confirmed unchanged since that verification by
-metadata (see "Safety invariant" for the exact guarantee). Two new job
+metadata (see "Safety invariant" for the exact guarantee). Three new job
 types plus an import-page UI section. No changes to the import pipeline
 itself.
 
@@ -173,9 +173,9 @@ verified files** button opens a confirmation dialog that states plainly:
 - Deletion is permanent — memory cards have no trash.
 - After deletion, the archive holds the only copy of these photos.
 - What "verified" means: each file's archive copy passed a checksum check
-  (at import or integrity audit) and is confirmed unchanged since by size
-  and timestamp; the card copy itself is re-checksummed at the moment of
-  deletion. Archive bytes are not re-downloaded.
+  (at import, targeted cleanup verification, or integrity audit) and is
+  confirmed unchanged since by size and timestamp; the card copy itself is
+  re-checksummed at the moment of deletion. Archive bytes are not re-downloaded.
 
 Confirming starts the **delete** job with live per-file progress. The final
 summary reports exact counts: deleted, kept, skipped-because-changed, and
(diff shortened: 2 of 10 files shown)


jss367:
Went through the 5 non-outdated Codex threads against the current head (`e51638b3`) and pushed `4fb818d7`:

- **[P2 – card_cleanup.py:647](https://github.com/jss367/vireo/pull/1453#discussion_r3761155752) "Invalidate deleted entries before refreshing verification"** — **fixed here.** `delete_verified` unlinks card files but never rewrites the manifest, so its `deletable` entries sit on `manifest["entries"]` when `verify_manifest_archives` runs next. The recomputed totals then credited the deleted bytes to the deletable bucket, the manifest revision bumped, and a subsequent delete confirmation for the inflated set would clear the freshness gate. Added a pre-pass that drops `deletable` entries whose card path `lstat` `FileNotFoundError`; a transient `OSError` preserves the entry so a NAS blip cannot shrink the preview. Two tests cover the drop path (deleted card file → excluded from refreshed totals, revision still bumps) and the transient-error path (flaky `lstat` → entry preserved).

- **[P2 – card_cleanup.py:460](https://github.com/jss367/vireo/pull/1453#discussion_r3760467123) "Offer verification when an unchecked duplicate can qualify"** — already covered by the priority fix in `qualify_rows` (lines 395-402 / 401-402): whenever `saw_never_checked` is true, `KEEP_NOT_VERIFIED` overrides any specific reason set from an `ok` row that failed a later gate, so the entry lands in the bucket the verify filter reads.

- **[P1 – card_cleanup.py:579](https://github.com/jss367/vireo/pull/1453#discussion_r3760877406) "Re-stat the archive pathname before certifying it"** — already covered by commit `672a2c4a`: lines 600-615 `os.stat(archive_path)` after hashing and reject certification unless the pathname's dev/inode still match the fd's stat (plus the existing before/after fd checks).

- **[P2 – card_cleanup.py:672](https://github.com/jss367/vireo/pull/1453#discussion_r3761014635) "Keep transient failures eligible for retry"** — already covered by commit `e51638b3`: lines 659-670 in the post-verify pass preserve `KEEP_NOT_VERIFIED` from `qualify_rows` (viable unchecked row) instead of overwriting it with the run's transient `KEEP_ARCHIVE_UNREACHABLE`/`CHANGED`, so a targeted retry after the mount returns still finds pending work.

- **[P2 – card_cleanup.py:532](https://github.com/jss367/vireo/pull/1453#discussion_r3760660505) "Preserve unchecked status when the archive is offline"** — the DB persistence and retry-eligibility halves are addressed by commits `9495ddea` + `e51638b3`: `os.open` failure no longer calls `update_photo_hash_check`, so the row stays NULL and `qualify_rows` → post-pass keeps the entry in `KEEP_NOT_VERIFIED`. What remains is a display honesty question — `stats["unreadable"]` is bumped on unreachable paths, and the frontend rolls that into "N could not be verified". Splitting into a separate `unreachable` counter with distinct UI copy is worth doing but the change reaches the callout wording — leaving it as a follow-up rather than piggybacking on this PR unless you'd like it included here.

Pinned test suite: 2110 passed / 1 pre-existing failure unrelated to card cleanup (`test_failed_import_does_not_chain` — depends on `chmod 0` denying root, which this container's user can bypass; reproduces on the parent commit `e51638b3`).

---

jss367:
Addressed the three outstanding Codex findings on the current head across two commits:

- **[P2 – card_cleanup.py:478](https://github.com/jss367/vireo/pull/1453#discussion_r3761288235) "Revalidate changed entries before retaining them"** — fixed in `23b5344e` (parallel session). Pre-pass now also drops entries where `lstat` succeeds but size/mtime no longer match the manifest, and rejects symlink/non-regular replacements, matching what `delete_verified` would already skip.

- **[P2 – card_cleanup.py:603](https://github.com/jss367/vireo/pull/1453#discussion_r3761288241) "Stop retrying same-file archive aliases"** — fixed in `23b5344e` (parallel session). Verify now tracks a per-hash `all_unchecked_inside_source` flag and overrides `qualify_rows`' `KEEP_NOT_VERIFIED` with `KEEP_INSIDE_SOURCE` in the reclassification pass when every unchecked row terminally resolved to the card file itself, so the retry callout stops.

- **[P2 – app.py:20179](https://github.com/jss367/vireo/pull/1453#discussion_r3761344386) "Document the required delete manifest revision"** — fixed here (`b229fb61`). Updated `docs/superpowers/specs/2026-08-07-card-cleanup-design.md`: the delete-job prose and the endpoint list now state that the body is `{scan_job_id, manifest_revision: int}`, describe the `409 manifest_revision_mismatch` response, and note the explicit bool rejection (`isinstance(True, int)` is truthy). Doc-only change.

Pinned test suite is still green (2110+ pass, 1 pre-existing `chmod 0`-vs-root failure unrelated to this PR).

---

jss367:
🤖 No action taken — the comment is a status update, and both referenced commits (`23b5344e` for the two `card_cleanup.py` P2 fixes and `b229fb61` for the spec update) are already on this PR's head (`b229fb61`). Nothing outstanding for this routine to push.

---

jss367:
Addressed the one new Codex finding from the review of `dab3acfc` in `b176280`:

- **[P2 – card_cleanup.py:746](https://github.com/jss367/vireo/pull/1453#discussion_r3761540976) "Revalidate pending card entries before promoting them"** — fixed. The deletable pre-pass already revalidated card files against the manifest baseline, but pending `KEEP_NOT_VERIFIED` entries went straight into `qualify_rows`, which only looks at catalog rows. A card file replaced/resized/turned into a symlink between the scan and the refresh therefore landed in the refreshed deletable totals, even though `delete_verified` would skip it (`SKIP_CHANGED` / `SKIP_SYMLINK` / `SKIP_NOT_REGULAR`).

  Added `_card_gate_promotion_block` (parallel to `delete_verified`'s cheap card-side gates) and consult it before setting `bucket = "deletable"` on any pending entry in the promotion loop. On a gate miss the entry is reclassified as `kept` with the matching `SKIP_*` reason and `archive_path` is dropped. Chose the promotion-time check over extending the pre-pass so archive verification still runs and refreshes the catalog row's `hash_status` — the pending entry is disqualified from promotion but the archive-side work isn't wasted. Content drift with size/mtime intact is intentionally left to `delete_verified`'s full-hash pass; repeating it here would defeat the point of scoped verification.

  New test `test_scoped_verify_gates_pending_entry_before_promotion`: mtime-bumps a pending card file between scan and verify, then asserts the archive still verifies (`verified == 1`) but nothing is promoted (`unblocked_files == 0`, deletable totals stay at 0) and the entry is reclassified as `kept` with `SKIP_CHANGED`.

Pinned test suite is green (2110 passed / same pre-existing `test_failed_import_does_not_chain` failure that predates this PR — `chmod 0` cannot deny root in this container; reproduces on parent). Ruff clean.

---

jss367:
Reviewed the two new Codex P2 findings from the review of `b176280e` against the current head. Both had already been addressed in parallel by commit `2dbbfa13` ("Harden scoped cleanup manifest refresh"); pushed `bd4490be` + `21b6a804` as complementary reinforcement:

- **[P2 – card_cleanup.py:456](https://github.com/jss367/vireo/pull/1453#discussion_r3761655412) "Drop missing pending entries during preview refresh"** — fixed by `2dbbfa13`, which added a `SKIP_ALREADY_GONE` return to `_card_gate_promotion_block` (line 461-462) and a `missing_pending_entry_ids` drop pass after the promotion loop (line 774-802 + 848-852). A pending KEEP_NOT_VERIFIED entry whose card file is gone is dropped from the manifest after scoped verify runs, so the refreshed totals no longer credit the missing bytes to the kept bucket. The archive still gets hashed and the catalog row still updates, which was a deliberate design choice — the test `test_scoped_verify_drops_missing_pending_entry` asserts `verified == 1` for exactly that reason. My initial commit (`bd4490be`) short-circuited that flow with a pre-pass FileNotFoundError drop, which broke that test; `21b6a804` reverts the pre-pass extension so the drop happens in the promotion loop as intended.

- **[P2 – card_cleanup.py:820](https://github.com/jss367/vireo/pull/1453#discussion_r3761655420) "Preserve permanent unreadable verdicts in the preview"** — fixed by `2dbbfa13`, which added an `elif reason == KEEP_ARCHIVE_HASH_FAILED:` branch (line 832-838) so the promotion-loop override preserves `qualify_rows()`'s terminal verdict for a persisted-unreadable row instead of replacing it with the run's transient `failure_reasons[hash]`. `bd4490be` complements that from the other side: in the three scoped-verify branches that persist a row as `hash_status='unreadable'` (fstat failure after successful open, non-regular file, mid-read `compute_fd_hash` failure), `outcome_reason` was left at the initial `KEEP_ARCHIVE_UNREACHABLE` from line 582 rather than being set to reflect the persisted verdict. Setting `outcome_reason = KEEP_ARCHIVE_HASH_FAILED` in those three branches makes `failure_reasons[hash]` carry the persisted verdict directly, so the two defenses agree even when `qualify_rows`'s verdict is bypassed (e.g., if a subsequent scoped-verify iteration hits a different `qualify_rows` code path).

New regression tests in `bd4490be`:
- `test_scoped_verify_drops_pending_kept_entries_whose_card_files_are_gone` — mixed present + absent pending case, complementing `test_scoped_verify_drops_missing_pending_entry`'s single-entry case
- `test_scoped_verify_keeps_pending_entry_when_lstat_error_is_transient` — pins that a transient (non-FileNotFoundError) OSError on the card lstat does not drop the pending entry
- `test_scoped_verify_records_hash_failed_when_open_fd_hash_fails` — mid-read `compute_fd_hash` failure persists the row as `unreadable` and the refreshed manifest reason is `KEEP_ARCHIVE_HASH_FAILED`, not `KEEP_ARCHIVE_UNREACHABLE`

Pinned test suite: 2222 passed / 1 pre-existing `chmod 0`-vs-root failure that predates this PR (`test_failed_import_does_not_chain`, reproduces on parent). Ruff clean.

---

jss367:
Addressed the Codex review of `de46328096` in `09de8d0a` (rebased onto a parallel session's push).

- **[P1 – card_cleanup.html:977](https://github.com/jss367/vireo/pull/1453#discussion_r3761738764) "Keep deletion locked while reloading a mismatched revision"** — fixed here.

  The `manifest_revision_mismatch` branch called `cardCleanupSetBusy(false)` before kicking `cardCleanupLoadManifest()` off as fire-and-forget, so Scan/Verify/Delete were re-enabled in the gap before the reload response landed. A user who reopened the confirm dialog during that gap saw the pre-mismatch totals; once the async reload replaced `cardCleanupState.manifest` with the fresh one, confirming would read the new revision and the server would accept deletion of the newly promoted set that never appeared in the dialog.

  Await `cardCleanupLoadManifest` before unlocking. On success, the render inside the reload has already set Delete's disabled state and hint from the fresh totals; unlock with `keepDeleteState: true` so `busyRestore` does not overwrite them. On failure, the render never ran, so explicitly disable Delete and swap the stale "Deletion is starting…" hint for a "Reload the page or scan again before deleting" instruction — a stale preview must not be able to authorize a deletion.

  New pinning test `test_confirm_delete_awaits_manifest_reload_on_revision_mismatch` walks the mismatch branch: the reload is awaited, the unlock runs after the await and uses `keepDeleteState: true`, and the failure branch explicitly disables Delete with the recovery hint.

- **[P2 – card_cleanup.py:465](https://github.com/jss367/vireo/pull/1453#discussion_r3761655412) "Drop missing pending entries during preview refresh"** — already addressed in `bd4490be` (parallel session): the pre-pass now handles pending entries too, so `FileNotFoundError` on a pending card path drops it from the manifest, transient `OSError` preserves it, and post-rebase interaction with the existing SKIP_ALREADY_GONE handler was cleaned up in `21b6a804`.

- **[P2 – card_cleanup.py:842](https://github.com/jss367/vireo/pull/1453#discussion_r3761655420) "Preserve permanent unreadable verdicts in the preview"** — already addressed in `bd4490be`: the three scoped-verify branches that persist `hash_status='unreadable'` after opening the archive fd now set `outcome_reason = KEEP_ARCHIVE_HASH_FAILED`, so `failure_reasons[hash]` carries the terminal verdict and the promotion-loop override matches `qualify_rows()`'s reason (complementing the existing `elif reason == KEEP_ARCHIVE_HASH_FAILED` guard from 2dbbfa13).

Pinned test suite is green (2230 passed / 1 pre-existing unrelated `test_failed_import_does_not_chain` failure — `chmod 0` cannot deny root in this container; reproduces on parent). Card cleanup tests: 123 passed / 1 skipped. Ruff clean.

---

jss367:
🤖 No action taken — the comment is a status update, and both referenced commits (`09de8d0a` for the P1 `card_cleanup.html` fix, which is this PR's current head, and `bd4490be` for the two P2 `card_cleanup.py` fixes done in a parallel session) are already on the branch. Nothing outstanding for this routine to push.

---

jss367:
Addressed the one new Codex finding from the review of `21b6a8043d` in `6a97adf`:

- **[P2 – card_cleanup.py:610](https://github.com/jss367/vireo/pull/1453#discussion_r3761826357) "Surface failed external copies alongside unchecked aliases"** — fixed.

  Scenario: a hash has one external archive row already persisted as `corrupt`/`modified`/`unreadable` from a prior verify pass AND one same-source alias row still NULL. Scoped verify only processes the NULL alias, catches it via the fd's dev/inode, and adds the hash to `terminal_inside_source_hashes` (all unchecked rows terminally resolved inside-source). `qualify_rows` still returns `KEEP_NOT_VERIFIED` because the NULL alias wins its priority order, and the reclassification override then rewrote that to `KEEP_INSIDE_SOURCE` — telling the user "no independent archive exists" when in fact one was cataloged and hashed as corrupt, and hiding the Audit-page remedy the failed row needs.

  Fix: in the `elif reason == KEEP_NOT_VERIFIED:` branch, when the terminal-alias override fires, first check whether any row for this hash has a persisted check-failed `hash_status`. If so, prefer `KEEP_ARCHIVE_HASH_FAILED` (routes to Audit) over `KEEP_INSIDE_SOURCE`. Every inside-source / same-as-card gate in the verify loop short-circuits with `continue` before `update_photo_hash_check` runs, so any persisted non-None non-"ok" `hash_status` here is definitionally an external row — checking `rows` directly is enough; no separate tracking set is needed.

  New regression test `test_scoped_verify_terminal_alias_does_not_mask_persisted_external_failure` walks the scenario end-to-end: seeds an external row with `hash_status="corrupt"` plus a hardlink alias row, runs `verify_manifest_archives`, and asserts the refreshed manifest reason is `KEEP_ARCHIVE_HASH_FAILED` (not `KEEP_INSIDE_SOURCE`). Fails on the unmodified branch head; passes with the fix. The existing `test_scoped_verify_terminal_alias_promotes_to_inside_source` still passes — the alias-only case has no persisted external failure, so the override still lands on `KEEP_INSIDE_SOURCE` as before.

Pinned test suite: 2110 passed / 1 pre-existing unrelated `test_failed_import_does_not_chain` failure (`chmod 0` cannot deny root in this container; reproduces on parent). Card cleanup tests: 124 passed / 1 skipped. Ruff clean.

---

jss367:
🤖 No action taken — the comment is a status update, and the referenced commit `6a97adf` (the `card_cleanup.py:610` fix that prefers `KEEP_ARCHIVE_HASH_FAILED` over `KEEP_INSIDE_SOURCE` when a persisted external failure exists, plus the new `test_scoped_verify_terminal_alias_does_not_mask_persisted_external_failure` regression) is already this PR's current head. Nothing outstanding for this routine to push.

---

